PR #16, opened 2019-04-17
Viewer references

paulwarren-wk:
Incorporate taxonomy reference information into the viewer.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/inspector.js
+++ b/iXBRLViewerPlugin/viewer/src/js/inspector.js
@@ -108,14 +108,34 @@ Inspector.prototype.search = function (s) {
                 .text(r.fact.getLabel("std")) // + " (" + r.score + ")" );
                 .appendTo(row);
         }
-
     });
     viewer.highlightRelatedFacts($.map(results, function (r) { return r.fact } ));
-    
 }
 
 Inspector.prototype.updateCalculation = function (fact, elr) {
     $('.calculations .tree').html(this._calculationHTML(fact, elr));
+}
+
+Inspector.prototype._referencesHTML = function (fact) {
+    var c = fact.concept();
+    var a = new Accordian();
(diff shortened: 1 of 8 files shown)


aviary3-wk:
## Security Insights
<details>
<li>Watched keyword <code>.html(t</code> in <code>iXBRLViewerPlugin/viewer/src/js/inspector.js</code> line(s) ['304'] added</li>
</details>

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -281,6 +301,7 @@ Inspector.prototype.update = function () {
             );
         this._updateEntityIdentifier(fact);
         this.updateCalculation(fact);
+        $('div.references').html(this._referencesHTML(fact));
```
brettkail-wk: In general, our security team prefers if we avoid jQuery `.html`.  Is it reasonable to use `.clear().append(this._referencesHTML(fact))` or similar construct instead?

I'll admit I'm not overly familiar with jQuery, so I'm not quite following this: it looks like `Accordian.html` returns a jQuery object (not an HTML string), but it looks like jQuery `.html` only accepts HTML strings, so what does this actually do?  Does jQuery `.html` also accept jQuery objects, or is there some HTML stringifying that I'm missing?

paulwarren-wk: You're right.  I thought that _referencesHTML was returning HTML as the last line is return a.html(), but it's not - as you say, it's a jQuery object, and the docs don't suggest that this should work at all.
I've addressed this in 732b68e, and renamed the html() method to avoid confusion with the jquery .html() method.

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully. Reference section is now visible in Fact Details panel.

@Workiva/release-management-p

Issue #12, opened 2019-04-09
Issue with "make prod"


xbrles:
Dear all,

We have a problem at the moment we run "make prod", the process never finishes and there are no messages. 

We're using Windows 10 and the last version available of npm (6.9.0).

Executing "npm run test" bring some errors:

![image](https://user-images.githubusercontent.com/19322084/55820025-caae6600-5b02-11e9-8d3a-63707e445c31.png)

![image](https://user-images.githubusercontent.com/19322084/55820095-f7fb1400-5b02-11e9-89ec-7a8b6dbb6487.png)

Also attached the LOG of the execution of the "npm run test".

[2019-04-09T17_03_06_247Z-debug.log](https://github.com/Workiva/ixbrl-viewer/files/3060169/2019-04-09T17_03_06_247Z-debug.log)

Any support is very much appreciated.

Thanks in advance for your help!

--------------
Javier Mora
XBRL Spain
info@xbrl.es
--------------

paulwarren-wk:
bc3aa0a9835395d40ba15b03151e43da23320d0a should hopefully fix the issue with the tests.  

I'm not sure what the issue is with make prod.  Does "make dev" work any better?

xbrles:
Hi Paul,

Thanks for your help! Let me explain you step by step what I'm doing.

**STEP 1.** After installing npm, I've run "npm install" with this result:

![image](https://user-images.githubusercontent.com/19322084/56461648-90707e80-63b6-11e9-97d6-e780b28943f7.png)

Please, let me know if I need to fix something related to the warnings.

**STEP 2.** After your last commit, I've successfully run the javascript unit tests with "npm run test"

![image](https://user-images.githubusercontent.com/19322084/56461663-e3e2cc80-63b6-11e9-941e-21f21624fd32.png)

**STEP 3.** Both commands: "make prod" and "make dev" have the same result, the blinking underscore symbol as the process never finishes and no error messages related.

![image](https://user-images.githubusercontent.com/19322084/56461668-f8bf6000-63b6-11e9-870e-40c52927fafe.png)

Any ideas of how can I solve this issue?

Thanks in advance for your help!

-----------------------
Javier Mora
XBRL Spain
info@xbrl.es
-----------------------

paulwarren-wk:
Can you try doing:

```
cd iXBRLViewerPlugin\viewer
npx webpack --config webpack.dev.js --optimize-minimize
```

xbrles:
[...]

PR #18, opened 2019-04-18
Timezone fix

paulwarren-wk:
Fix the issue with date tests failing mentioned in issue #12.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/util.js
+++ b/iXBRLViewerPlugin/viewer/src/js/util.js
@@ -32,9 +32,7 @@ export function momentToHuman(d, adjust) {
         }
         return d.format('D MMM Y');
     }
-    else {
-        return d.format("D MMM Y HH:mm:ss");
-    }
+    return d.format("D MMM Y HH:mm:ss");
 }
 
 /*
--- a/iXBRLViewerPlugin/viewer/src/js/util.js
+++ b/iXBRLViewerPlugin/viewer/src/js/util.js
@@ -19,7 +19,11 @@ import moment from "moment";
  * Takes a moment.js oject and converts it to a human readable date, or date
  * and time if the time component is not midnight.  Adjust specifies that a
  * date (but not a date time) should be shown as the day before.  This is to
- * satisfy the convention of describing durations using inclusive dates.
+ * satisfy the human convention of describing durations using inclusive dates.
+ *
+ * i.e. 2018-01-01T00:00:00 to 2019-01-01T00:00:00 is described as
+ *      "Jan 1 2018 to Dec 31 2018"
+ *
  */
 export function momentToHuman(d, adjust) {
     if (d.hours() + d.minutes() + d.seconds() == 0) { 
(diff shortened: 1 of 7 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -13,25 +13,38 @@
...
 import dateFormat from "dateformat"
+import moment from "moment";
 
-export function isodateToHuman(d, adjust) {
-    if (d.getUTCHours() + d.getUTCMinutes() + d.getUTCSeconds() == 0) { 
+/* 
+ * Takes a moment.js oject and converts it to a human readable date, or date
+ * and time if the time component is not midnight.  Adjust specifies that a
+ * date (but not a date time) should be shown as the day before.  This is to
+ * satisfy the convention of describing durations using inclusive dates.
```
brettkail-wk: Describing it in terms of "inclusive" seems odd to me since the adjustment is needed because duration end dates are exclusive (and for instances, which aren't inclusive or exclusive).  Perhaps "This is to satisfy the convention of describing 24:00:00 as 00:00:00 of the following day for instants and duration end dates."

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -73,10 +86,26 @@ export function wrapLabel(str, maxwidth){
...
+export function xbrlDateToMoment(dateString) {
+    /* If the string has something after the date part other than a time part,
+     * insert a time part of 'T00:00:00'
+     *
+     * i.e. 2010-01-01Z => 2010-01-01T00:00:00Z
+     *
+     */
+    dateString = dateString.replace(
+        /^(\d{4,}-\d{2}-\d{2})(?!T|$)/, 
+        function(match, $1) { return $1 + 'T00:00:00' }
```
brettkail-wk: I don't understand how this works for instants or duration end dates.  If the value is `2019-03-31` in iXBRL, this will become `2019-03-31T00:00:00`, and then `momentToHuman` will render it as `2019-03-30`.  I believe this function similarly needs `adjust`, and then if there is no "T" (so a plain date) and `adjust` (so end date or instant), then it should `.add(1, 'day')` per
> A date, with no time part, in the endDate or instant element is defined to be equivalent to specifying a dateTime of the same date plus P1D and with a time part of T00:00:00.

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -13,25 +13,38 @@
...
+export function momentToHuman(d, adjust) {
+    if (d.hours() + d.minutes() + d.seconds() == 0) { 
         if (adjust) {
-            d.setDate(d.getDate() - 1);
+            d = d.clone().subtract(1, 'day');
         }
-        return dateFormat(d,"d mmm yyyy");
+        return d.format('D MMM Y');
     }
     else {
```
derekgengenbacher-wf: You can remove this else statement and just use the return from line 32.

paulwarren-wk: Done in 470bd1b

iXBRLViewerPlugin/viewer/src/js/util.js
paulwarren-wk: Internally, the viewer works with full date times, so an end date for one period and the start date for the next are the same time.  This comment is referring to the fact that the "human" way to describe the year from 2018-01-01T00:00:00 to 2019-01-01T00:00:00 is "1st Jan 2018 to 31st Dec 2018" i.e. the first and last days included in the period (with the side effect that we describe the end of one period and the beginning of the next differently even thought they're the same point in time).  I've expanded the comment in 395114d0d946973bad5e0381a5d6d13fb3a778ff .

brettkail-wk: Ah, my misunderstanding.  Thanks for clarifying.

iXBRLViewerPlugin/viewer/src/js/util.js
paulwarren-wk: The trick is that we're not working with raw XBRL 2.1 dates.  We're working with dates from the JSON block which are truncated consistently to midnight at the start of the day.  This is consistent with the approach taken in xBRL-JSON: code and syntax should work consistently with datetimes, and the convention of adjusting instants and end dates should be applied purely with in the rendering layer (xBRL-JSON doesn't technically allow the removal of the time part at all).

brettkail-wk: I see, thanks for clarifying.  I did not appreciate the difference in processing and likely would have been stuck for quite some time, so thanks for fixing this issue :-).

derekgengenbacher-wf:
QA +10 Unit Tests pass successfully. Builds successfully. Downloaded and smoke tested locally. Dates come across correctly now.

@Workiva/release-management-p

PR #20, opened 2019-04-19
XHTML support

paulwarren-wk:
Make the viewer work when treated as XHTML.  iXBRL filings are often served with an XHTML extension or content type, which will put the browser into standards compliant rendering mode, and enforce XML validity.  

For most filings the rendering mode makes little or no difference, but for others it can substantially affect the rendering.

This PR makes the viewer itself XHTML valid, which means that it can be served using an XHTML extension/mime-type, with the result that the iXBRL file will also be treated as XHTML.

The SEC serves iXBRL with a .htm extension, so it is treated as HTML, but UK Companies House use .xhtml.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/html/inspector.html
+++ b/iXBRLViewerPlugin/viewer/src/html/inspector.html
@@ -14,13 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div id="ixv">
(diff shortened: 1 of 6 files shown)


aviary-wf:
## Security Insights

## Action Items
---

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully. Downloaded and smoke tested locally with an xhtml instance doc successfully.

@Workiva/release-management-p

PR #21, opened 2019-04-22
XMAS-3842 - Fix for facts with HTML content

paulwarren-wk:
If a fact appears to contain escaped HTML tags, show just the text content of that HTML, rather than the HTML source.  This fixes XMAS-3842.

A fact is determined to contain HTML if it starts with optional whitespace, followed by "<", QName, then ">", "/>" or space.

Ideally we'd check the fact datatype too, but we don't currently have access to datatype information in the viewer, and the chance (and downside) of a false positive seems minimal.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/fact.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.js
@@ -21,6 +21,7 @@ import $ from 'jquery'
 
 export function Fact(report, factId) {
     this.f = report.data.facts[factId];
+    this._ixData = report.getIXDataForFact(factId);
     this._report = report;
     this.id = factId;
 }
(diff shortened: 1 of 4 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -80,6 +80,22 @@ Fact.prototype.readableValue = function() {
             v = formatNumber(v,d) + " " + this.unit().valueLabel();
         }
     }
+    /* If the text value appears to start with an HTML tag, treat as HTML and
+     * use the text content as the value.  It would be better to apply this
+     * based on data type, but we don't currently have access to that in the
+     * report data.
+     */
+    else if (v.match(/^[\s\u00a0]*<([a-zA-Z_][\w.-]*:)?[a-zA-Z_][\w.-]*(\s|>|\/>)/)) {
```
brettkail-wk: It seems like this should be based on `<ix:nonNumeric escape="true">`.  If that attribute isn't currently available in the JSON, it seems like we should make it available.

paulwarren-wk: Good point!  I'd completely forgotten that the fact that we're coming from iXBRL means that we have the escaped attribute, which actually preferable to relying on a datatype and heuristic.  It should be possible to have the JS pull this straight from the iXBRL, rather than adding it to the JSON.

brettkail-wk:
@paulwarren-wk FYI, you have merge conflicts and need to merge/rebase.

iXBRLViewerPlugin/viewer/src/js/fact.js
paulwarren-wk: This is done in 163e12f

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully. Downloaded and tested locally. The Fact Details no longer shows html for the fact value for the table tagged with xbrl in the provided doc.

@Workiva/release-management-p

PR #29, opened 2019-05-21
XMAS-4131 Allow postMessage() events to navigate to a Fact by identifier

danaoredson-wf:
Listen to `postMessage()` in order for the ixbrl viewer to be driven by consumers without requiring consumers to have direct access to the javascript source code.

This adds handling for showing a Fact by identifier.

It does this by listening to `window.onMessage`. Integrators can call window `postMessage(msg, origin)`.

The format of this message is JSON. The only handled message introduced in this PR is of the following format:
```
{
  "task": "SHOW_FACT",
  "factId": "the fact id"
}
```

Invalid JSON message will throw an exception
A JSON message that does not define a 'task' attribute or does not have a task value of "SHOW_FACT" will be ignored.
An invalid factId will be ignored.

This can be tested [by downloading this zip](https://drive.google.com/open?id=1nJrbVuhVI3wfpaYMZJ5OAy4Q14P5vy57), generating ixbrlviewer.dev.js from this repo and replacing the file in the zip folder. Then open blackline-report.html.  You should then be able to regression-test the viewer itself there, plus click on Fact lines that are Updated or Added, and it *should* scroll to and select the corresponding Fact in the ixbrl viewer.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -226,9 +226,9 @@ Viewer.prototype.showAndSelectFact = function (fact) {
 }
 
 Viewer.prototype.showAndSelectFactById = function (factId) {
-    let fact = this.elementForFactId(factId);
-    if(fact) {
-        this.showAndSelectElement(fact);
+    let element = this.elementForFactId(factId);
+    if(element) {
+        this.showAndSelectElement(element);
     }
 }
 
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -145,7 +145,7 @@ Viewer.prototype.handleMessage = function (event) {
     var jsonString = event.originalEvent.data;
     var data = JSON.parse(jsonString);
 
-    if(data.task == 'SHOW_FACT') {
+    if (data.task == 'SHOW_FACT') {
         var factId = data.factId;
         this.showAndSelectFactById(factId);
     }
@@ -226,9 +226,9 @@ Viewer.prototype.showAndSelectFact = function (fact) {
 }
 
 Viewer.prototype.showAndSelectFactById = function (factId) {
-    let element = this.elementForFactId(factId);
-    if(element) {
-        this.showAndSelectElement(element);
+    let fact = this.elementForFactId(factId);
+    if (fact) {
+        this.showAndSelectElement(fact);
     }
 }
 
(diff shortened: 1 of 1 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -205,6 +225,13 @@ Viewer.prototype.showAndSelectFact = function (fact) {
     this.showAndSelectElement(this.elementForFact(fact));
 }
 
+Viewer.prototype.showAndSelectFactById = function (factId) {
+    let fact = this.elementForFactId(factId);
```
danaoredson-wf: oops, this variable should really be `element`. Will change.

danaoredson-wf: addressed

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -135,9 +135,25 @@ Viewer.prototype._bindHandlers = function () {
     $('#iframe-container .zoom-in').click(function () { viewer.zoomIn() });
     $('#iframe-container .zoom-out').click(function () { viewer.zoomOut() });
 
+    // Listen to messages posted to this window
+    $(window).on("message", function(e) { viewer.handleMessage(e) });
```
brettkail-wk: Perhaps `$(window).on("message", viewer.handleMessage);`?

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -135,9 +135,25 @@ Viewer.prototype._bindHandlers = function () {
...
+
     TableExport.addHandles(this._contents, this._report);
 }
 
+Viewer.prototype.handleMessage = function (event) {
+    var jsonString = event.originalEvent.data;
+    var data = JSON.parse(jsonString);
+
+    if(data.task == 'SHOW_FACT') {
```
brettkail-wk: Nit: repo not fully consistent, but space in `if (` seems 10x more prevalent.

iXBRLViewerPlugin/viewer/src/js/viewer.js
danaoredson-wf: I had tried that already and it doesn't work, as `this` is not populated in `handleMessage` when invoked that way, and so felt out of scope for this change.

iXBRLViewerPlugin/viewer/src/js/viewer.js
danaoredson-wf: nit addressed

iXBRLViewerPlugin/viewer/src/js/viewer.js
brettkail-wk: Oops, ok.  I don't do JavaScript often enough to remember the nuances of `this`.

danaoredson-wf: I suspect this is also the reason why Paul handle the zoom in/out click handling using anonymous lambda functions above.

brettkail-wk:
QA +1 Downloaded build artifact .tgz, extracted ixbrlviewer.js, downloaded referenced example zip, replaced ixbrlviewer.dev.js with build artifact, and clicking fact scrolls to item + expands fact as expected.
@Workiva/release-management-p

PR #34, opened 2019-05-27
Show all facts when selecting a nested tag

paulwarren-wk:
If a piece of text has been tagged more than once, clicking on it will now show all facts which it forms a part of in an accordian.

The code will generate a minimally unique label for each fact in the set to use as a title for the accordian.  It runs through each dimension on the facts, and adds the value for that dimension for all facts if there are any differences between the facts.  It does this as many times as necessary to get unique labels for each fact.  The first dimension it tries is "concept", an the second is "period".  This means that for common cases, the facts will be labelled by just concept, or just period.

The use of the accordian here is perhaps a little suspect, as the subsequent sections (references and calculations) are linked to the individual facts, and so will change if you switch between the different cards in the accordian.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/factset.js
+++ b/iXBRLViewerPlugin/viewer/src/js/factset.js
@@ -65,11 +65,13 @@ FactSet.prototype.minimallyUniqueLabel = function(fact) {
                 for (var i = 0; i < this._facts.length; i++) {
                     var fid = this._facts[i].id;
                     var l = allLabels[fid][j];
+                    var ul = uniqueLabels[fid] || [];
                     if (l !== null) {
-                        uniqueLabels[fid] = (uniqueLabels[fid] === undefined ? l : uniqueLabels[fid] + ", " + l);
+                        ul.push(l);
                     }
-                    if (uniqueLabels[fid] !== undefined) {
-                        uniqueLabelsByLabel[uniqueLabels[fid]] = true;
+                    if (ul.length > 0) {
+                        uniqueLabels[fid] = ul;
+                        uniqueLabelsByLabel[ul.join(", ")] = true;
                     }
                 } 
                 /* We have as many different labels as facts - we're done */
@@ -84,11 +86,13 @@ FactSet.prototype.minimallyUniqueLabel = function(fact) {
         if (Object.keys(uniqueLabels).length < this._facts.length) {
             for (var i = 0; i < this._facts.length; i++) {
                 var fid = this._facts[i].id;
-                uniqueLabels[fid] = uniqueLabels[fid] ? allLabels[fid][0] + ', ' + uniqueLabels[fid] : allLabels[fid][0];
+                var ul = uniqueLabels[fid] || [];
+                ul.unshift(allLabels[fid][0]);
+                uniqueLabels[fid] = ul;
             }
         }
 
         this._minimallyUniqueLabels = uniqueLabels;
     }
-    return this._minimallyUniqueLabels[fact.id];
+    return this._minimallyUniqueLabels[fact.id].join(", ");
 }
--- a/iXBRLViewerPlugin/viewer/src/js/factset.js
+++ b/iXBRLViewerPlugin/viewer/src/js/factset.js
@@ -1,0 +1,71 @@
+// Copyright 2019 Workiva Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export function FactSet(facts) {
+    this._facts = facts;
+}
+
+FactSet.prototype._allDimensions = function() {
+    var dims = {};
+    for (var i = 0; i < this._facts.length; i++) {
+        var dd = Object.keys(this._facts[i].dimensions());
+        for (var j = 0; j < dd.length; j++) {
+            dims[dd[j]] = 1;
+        }
+    }
+    return Object.keys(dims);
+}
+
+FactSet.prototype.minimallyUniqueLabel = function(fact) {
+    if (!this._minimallyUniqueLabels) {
+        var allLabels = {};
+        var allDims = this._allDimensions();
+        /* Assemble a map of arrays of all aspect labels for all facts, in a
+         * consistent order */
+        for (var i = 0; i < this._facts.length; i++) {
+            var f = this._facts[i];
+            allLabels[f.id] = [ f.getLabel(), f.period().toString() ];
+            for (var j = 0; j < allDims.length; j++) {
+                var dd = f.dimensions()[allDims[j]];
+                allLabels[f.id].push(dd ? dd.valueLabel : "");
+            }
+        }
+        /* Iterate each aspect label and compare that label across all facts in
+         * the set */
+        var uniqueLabels = {};
+        for (var j = 0; j < allDims.length + 2; j++) {
+            var labelMap = {};
+            for (var i = 0; i < this._facts.length; i++) {
+                labelMap[allLabels[this._facts[i].id][j]] = true;
+            }
+
+            /* We have at least some differences, so include this label */
+            var uniqueLabelsByLabel = {};
+            if (Object.keys(labelMap).length > 1) {
+                for (var i = 0; i < this._facts.length; i++) {
+                    var fid = this._facts[i].id;
+                    var l = allLabels[fid][j];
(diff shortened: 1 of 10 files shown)


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/factset.js
```diff
@@ -0,0 +1,84 @@
...
+export function FactSet(facts) {
+    this._facts = facts;
+}
+
+FactSet.prototype._allDimensions = function() {
+    var dims = {};
+    for (var i = 0; i < this._facts.length; i++) {
+        var dd = Object.keys(this._facts[i].dimensions());
+        for (var j = 0; j < dd.length; j++) {
+            dims[dd[j]] = 1;
```
brettkail-wk: Nit: `true`?

iXBRLViewerPlugin/viewer/src/js/factset.js
```diff
@@ -0,0 +1,84 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export function FactSet(facts) {
+    this._facts = facts;
+}
+
+FactSet.prototype._allDimensions = function() {
```
brettkail-wk: Suggest `/* Returns the dimension names relevant to some fact in the set. */`?

iXBRLViewerPlugin/viewer/src/js/factset.js
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            /* We have at least two different labels, so include this label */
+            var uniqueLabelsByLabel = {};
+            haveEmptyLabels = false;
+            if (Object.keys(labelMap).length > 1) {
+                for (var i = 0; i < this._facts.length; i++) {
+                    var fid = this._facts[i].id;
+                    var l = allLabels[fid][j];
+                    uniqueLabels[fid] = (uniqueLabels[fid] === undefined ? l : uniqueLabels[fid] + ", " + l);
```
brettkail-wk: Perhaps `allUniqueLabels`, local `uniqueLabels = []`, and `uniqueLabels.join(", ")` to avoid the repeated object key lookup + string concat.

iXBRLViewerPlugin/viewer/src/js/factset.js
```diff
@@ -0,0 +1,84 @@
...
+                    if (uniqueLabels[fid] === "") {
+                        haveEmptyLabels = true;
```
brettkail-wk: Should this be outside the loop?  What is the intended formatting if one value is empty but another is not?  If the concepts are all the same, will the duplicate concept not be included twice?

iXBRLViewerPlugin/viewer/src/js/factset.js
```diff
@@ -0,0 +1,84 @@
...
+            for (var j = 0; j < allDims.length; j++) {
+                var dd = f.aspects()[allDims[j]];
+                allLabels[f.id].push(dd ? dd.valueLabel() : "");
+            }
+        }
+        /* Iterate each aspect label and compare that label across all facts in
+         * the set */
+        var uniqueLabels = {};
+        var haveEmptyLabels;
+        for (var j = 0; j < allDims.length + 2; j++) {
```
brettkail-wk: This `+ 2` looks wrong.  Perhaps leftover from before `allDims` contained `["c", "p"]` also?  Remove the `+ 2` and rename that var to `labelKeys` or something?

iXBRLViewerPlugin/viewer/src/js/factset.js
```diff
@@ -0,0 +1,84 @@
...
+            var labelMap = {};
+            for (var i = 0; i < this._facts.length; i++) {
+                labelMap[allLabels[this._facts[i].id][j]] = true;
+            }
+
+            /* We have at least two different labels, so include this label */
```
brettkail-wk: Either this comment is misplaced, or I don't understand it.

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -367,8 +389,17 @@ Inspector.prototype.update = function () {
     this.updateURLFragment();
 }
 
-Inspector.prototype.selectFact = function (id) {
+Inspector.prototype.selectFact = function (id, factIdSet) {
```
brettkail-wk: This seems like an odd signature.  Add a second `selectFacts(ids)` and then change this one to call that one with `selectFact(id)`?

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -299,52 +300,69 @@ Inspector.prototype._updateEntityIdentifier = function (fact) {
...
+                inspector._currentFact = inspector._report.getFactById(id)
+                inspector._viewer.highlightFact(inspector._currentFact); 
+                inspector.update();
+            }, 
+            alwaysOpen: true,
+            dissolveSingle: true,
+        });
+        var fs = new FactSet(this._currentFactSet);
+        $.each(inspector._currentFactSet, function (i, fact) {
+            var factHTML = $(require('../html/fact-details.html'));
```
brettkail-wk: Is this file missing from the PR?

paulwarren-wk: Oops.  Added in dd1783a.

iXBRLViewerPlugin/viewer/src/js/factset.js
paulwarren-wk: Yep - should be within the "if".  Have hopefully improved the wording too (d617a3f)

iXBRLViewerPlugin/viewer/src/js/factset.js
paulwarren-wk: I think this was correct.  

If all concepts are the same, "concept" won't be added to the label as part of establishing unique labels, so there's no risk of it being duplicated.  This could have occurred if one of the concepts didn't actual have a label, but I've now fixed it to fall back on QName if there's no label, so concepts will always have a label.

The case that this code is intended to guard against is where you have, say, Revenue (Total) and Revenue (Europe).  If total is a default dimension value, the unique label code will give you "" and "Europe" as labels.  In this case we add back Concept as a label to get "Revenue" and "Revenue, Europe".

I've refactored this in order to handle the case of feeding it fully duplicate facts, and I think the end result is cleaner anyway.  Labels now don't get added to "uniqueLabels" until they've actually got some content, and then we just count how many we've got at the end to check that each fact has one (32ae878).

iXBRLViewerPlugin/viewer/src/js/factset.js
paulwarren-wk: 73a5aa8

iXBRLViewerPlugin/viewer/src/js/factset.js
paulwarren-wk: Improved in d617a3f

iXBRLViewerPlugin/viewer/src/js/factset.js
paulwarren-wk: I'm not sure if this is exactly what you were suggesting, but I've refactored it in 840c3b4

iXBRLViewerPlugin/viewer/src/js/factset.js
paulwarren-wk: Good catch, and yes that was where it came from.  I've renamed to allAspects in 4ccd1f7 .

iXBRLViewerPlugin/viewer/src/js/inspector.js
paulwarren-wk: The issue is that both id and factIdSet can be provided: the former is the specific fact to be selected, and the latter is a list of all facts that the click could correspond to (when you have nested facts).

As it stands, the selected fact will always be the first fact in the list, as this is 

I've attempted to clarify in dcb4524.

brettkail-wk: That helped, thanks.

iXBRLViewerPlugin/viewer/src/js/factset.js
```diff
@@ -0,0 +1,98 @@
...
+                /* We have as many different labels as facts - we're done */
+                if (Object.keys(uniqueLabelsByLabel).length == this._facts.length) {
+                    break;
+                }
+            }
+        }
+
+        /* If any label is empty, add the concept label onto the start of all
+         * of them */
+        if (Object.keys(uniqueLabels).length < this._facts.length) {
```
brettkail-wk: I think my misunderstanding [here](https://github.com/Workiva/ixbrl-viewer/pull/34#discussion_r288124188) is overloading of "label" in this function to mean both "single aspect value for a single fact" and "concatenation of all non-redundant aspect values for a single fact".  I was misreading this comment as "if any aspect value for any fact", but you mean "if any fact has either empty or redundant values for all aspects", right?  And since concept is a required aspect, the only way to get here is if all facts have the same concept but no value (or no non-redundant values) for other aspects, so we can just add the concept to the front.

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully. Smoke tests passed successfully.

@Workiva/release-management-p

PR #50, opened 2019-07-14
Inline XBRL document set support

paulwarren-wk:
Add support for Inline XBRL document sets.

When given an iXBRL document set as input, the plugin will embed the viewer into the first file, which will then load the other documents in the set into a set of tabbed iframes.  All files in the set get processed to ensure that all facts have ID attributes.

The viewer will automatically switch tabs when needed to focus on a fact. 

Due to security restrictions, the viewer does not work on document sets when working from local files (`file:` URLs) in Chrome.  A popup dialog is shown in this case.  It works fine in Firefox, or when loaded from `http` URLs.

A sample ixds is included as "ixds-test".

The sample document system has been reworked to support processing sets of documents.


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/index.js
```diff
@@ -55,24 +55,40 @@ $(function () {
...
+            $('#ixv .loader .text').text("Error: Could not find viewer data");
+            $('#ixv .loader').removeClass("loading");
+            return;
+        }
+        var report = new iXBRLReport(JSON.parse(taxonomyData));
+        if (report.isDocumentSet()) {
+            $('#ixv .ixds-tabs').show();
+            var ds = report.documentSetFiles();
+            for (var i = 1; i < ds.length; i++) {
+                var iframe = $("<iframe src=\"" + ds[i] + "\" />").appendTo("#ixv #iframe-container");
```
brettkail-wk: Use `.attr` to avoid HTML injection?

iXBRLViewerPlugin/viewer/src/js/index.js
```diff
@@ -55,24 +55,40 @@ $(function () {
...
+                iframes = iframes.add(iframe);
+            }
+        }
 
         /* Poll for iframe load completing - there doesn't seem to be a reliable event that we can use */
```
brettkail-wk: Out of curiosity, why doesn't `$(...).ready(...)` work?

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -84,9 +107,11 @@ Viewer.prototype._preProcessiXBRL = function(n, inHidden) {
         $(n).wrap(wrapper);
         node = $(n).parent();
     }
-    node.addClass("ixbrl-element").data('ivid',n.getAttribute("id"));
+    var id = n.getAttribute("id");
+    node.addClass("ixbrl-element").data('ivid',id);
+    this._factsByDocument[id] = doc;
```
brettkail-wk: This seems confusingly named.  Rename to `_documentsByFactId`?

Well, it took me quite some time to understand that `doc` is actually an index, not a reference to a jQuery/DOM object.  Perhaps `rename `doc` to `docIndex` throughout and possibly name `_docIndexesByFactId`?  Alternatively, perhaps some comments would help.

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -91,6 +90,7 @@ Inspector.prototype.handleMessage = function (event) {
     var data = JSON.parse(jsonString);
 
     if (data.task == 'SHOW_FACT') {
+        this.showDocumentForFactId(data.factId);
         this.selectFact(data.factId);
```
brettkail-wk: Does `handleFactDeepLink` need to be similarly updated?  Perhaps both should be using a common function (`showAndSelectFact`?) and the `showDocumentForFactId` logic should be there?

iXBRLViewerPlugin/viewer/src/js/index.js
```diff
@@ -51,33 +50,55 @@ function getTaxonomyData() {
     return null;
 }
 
+function checkDocumentSetBrowserSupport() {
+    if (document.location.protocol == 'file:' && window.chrome) {
```
derekgengenbacher-wf: Why does Chrome not support this?

brettkail-wk: Chrome is very strict about loading local resources (even more strict than Firefox).  I don't know if it's the same issue, but [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=40787) has some details about AJAX.

iXBRLViewerPlugin/viewer/src/js/index.js
paulwarren-wk: Good point. Done in b0ef76c .

iXBRLViewerPlugin/viewer/src/js/inspector.js
paulwarren-wk: Actually, this was a bug introduced by merging in the refactoring done for ix:continuation (c55557c).  selectFact already switches tabs if needed.  Fixed in c3ad9f5.

iXBRLViewerPlugin/viewer/src/js/index.js
paulwarren-wk: The viewer relies on running JS across iframes (i.e. the parent document modifies the contents of the iframes).  This is OK if the iframes are the same origin, but Chrome treats a pair of file URLs as cross-origin.  In single document mode, the iframe is populated dynamically from the same document, which apparently is OK.

We discussed adding a mode to the plugin to have it generate the viewer as a single (potentially very large) HTML file, and then populating all iframes dynamically.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: Yes - that's much clearer.  Done in 0cb3e95.

iXBRLViewerPlugin/viewer/src/js/index.js
paulwarren-wk: I'm not entirely sure, but I've just tried it and it doesn't seem to.  I remember searching when I originally wrote this and confirming that there wasn't an easy way.  I'm not sure if it was specific to the fact that we dynamically populate the iframe.

derekgengenbacher-wf:
This code all looks good, but can you update the documentation with how you would run the viewer via command line with both the viewer plugin and the ixds plugin please?

davecasleton-wf:
@paulwarren-wk See Derek's comment above.

samples/Makefile-src
```diff
@@ -12,29 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-HTM_TARGETS = $(patsubst %.htm, ../../viewers/%-viewer.htm, $(wildcard *.htm))
-HTML_TARGETS = $(patsubst %.html, ../../viewers/%-viewer.html, $(wildcard *.html))
-XHTML_TARGETS = $(patsubst %.xhtml, ../../viewers/%-viewer.xhtml, $(wildcard *.xhtml))
-
 # eval followed by space is a silly trick to match a space
-PACKAGES = $(subst $(eval) ,|,$(wildcard ../../packages/*.zip))
```
brettkail-wk: I guess the `eval` comment is no longer relevant.

paulwarren-wk: Yep - fixed in 80795da

derekgengenbacher-wf:
@paulwarren-wk When running the following command:

`./samples/build-viewer.py --viewer-url ixbrlviewer.js --out out-dir ~/Downloads/mdi/`

I am generating the following error:

```
Traceback (most recent call last):
  File "./samples/build-viewer.py", line 83, in <module>
    cntlr.createViewer(f, outPath = args.out, scriptUrl = args.viewer_url)
  File "./samples/build-viewer.py", line 58, in createViewer
    viewer.save(outPath)
AttributeError: 'lxml.etree._ElementTree' object has no attribute 'save'
```

The mdi directory contains a mulit-instance document that we have been using for testing. There are two html files in it.

paulwarren-wk:
> @paulwarren-wk When running the following command:
> 
> `./samples/build-viewer.py --viewer-url ixbrlviewer.js --out out-dir ~/Downloads/mdi/`
> 
> I am generating the following error:
> 
> ```
>     viewer.save(outPath)
> AttributeError: 'lxml.etree._ElementTree' object has no attribute 'save'
> ```

That looks like it's finding the wrong (master rather than branch) version of iXBRLViewerPlugin.  What's your PYTHONPATH set to?

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully. Locally tested in Chrome and Firefox and was able to successfully generate the browser warning in both.

@Workiva/release-management-p

PR #56, opened 2019-08-10
Set iframe height to 0 rather than hiding to improve performance.

paulwarren-wk:
This fixes issue 55 by reducing the height of iframes on inactive tabs to zero rather than actually hiding them.  This appears to give significantly better performance.

I've not noticed any negative side-effects, but I suspect that  memory usage may be worse on document sets with multiple large documents.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -387,8 +387,8 @@ Viewer.prototype.selectDocument = function (docIndex) {
         .eq(docIndex)
         .addClass("active");
     this._iframes
-        .hide()
+        .height(0)
         .eq(docIndex)
-        .show();
+        .height("100%");
     this._setTitle(docIndex);
 }


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -387,8 +387,8 @@ Viewer.prototype.selectDocument = function (docIndex) {
         .eq(docIndex)
         .addClass("active");
     this._iframes
-        .hide()
+        .height(0)
         .eq(docIndex)
-        .show();
+        .height("100%");
```
brettkail-wk: Changing the height feels weird.  Searching around suggests jQuery hide()/show() are slow (or at least were a few years ago), and all the posts I've found suggest that manipulating the CSS display property directly should be faster.  Did you try `.css("display", "none")` and `.css("display", "")`?  That seems preferable to adjusting the height if it works, but if that is also slow, I can go along with the height adjustment here.

paulwarren-wk: I tried manipulating the CSS properties directly via the DOM inspector in Chrome and that was slow. It does seem that hiding and showing via CSS causes the browser to re-render (also tried moving the hide/show onto different elements but with no change). I've just tried using .css() and it seems to behave the same.

I agree that setting height is a hack, but I can't find a better way.  FWIW, FF and Chrome seem to behave the same way on this.

brettkail-wk: A comment might be good to explain why we're doing something non-obvious.

paulwarren-wk: Good point - 6af8f9c

PR #58, opened 2019-08-25
Sort files in ixds alphabetically when converting samples

paulwarren-wk:
File order is significant when assembling viewers for document sets because a) the first file is treated as the "entry point" that pulls in all other files in the set and b) it controls the order of the tabs.

The is no standard iXBRL manifest that provides an ordering.  Previously sample files were being created in whatever order Python's glob() function was returning them, making it OS-dependent which order the set was assembled.  This PR sorts the files alphabetically, which at least gives us stable ordering and actually seems to give a good order for SEC document sets (at least on the examples seen so far).


aviary3-wk:
## Security Insights

## Action Items
---

PR #63, opened 2019-09-23
Search and other UI improvements

paulwarren-wk:
This PR provides a new UI for searching, putting the search function on a separate page in order to declutter the fact inspector.  

More details about search results are shown (period and dimensions) to avoid multiple identical results.

Styling for "no fact selected" and "no search results" is also improved.


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/html/inspector.html
```diff
@@ -47,7 +47,7 @@ <h1 >Fact Properties</h1>
             </div>
           </div>
           <div class="search-entry">
-            <input id="ixbrl-search" type="text" ></input>
+            <input id="ixbrl-search" type="text" />
```
brettkail-wk: I don't understand this commit.  These should be equally valid from an XHTML perspective?

paulwarren-wk: Yes.  The issue is that they get pre-processed by webpack's html-loader.  `<input></input>`  is technically invalid under (non-XML) HTML5, so html-loader minimises it into `<input>`, which is XML invalid.  `<input />` is both HTML and XHTML valid, and as we have `keepClosingSlash` turned on, html-loader preserves it.

brettkail-wk: Ah, so "Fix HTML5 validity" might have been a better commit message.  Regardless, it all makes sense now, thanks.

paulwarren-wk: Well, the issue that it fixed was that if you created a viewer with a .xhtml extension, the viewer failed to load because it wasn't XML valid.  But more words were probably needed, as it took me 5 minutes to remind myself of the reason!

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Pulled locally and verified search and fact details are on different tabs now. Search results also include dates and dimensions. 

@Workiva/release-management-p

PR #112, opened 2020-06-13
Add support for wider narrower (anchoring) relationships

paulwarren-wk:
Add support for anchoring relationships:

* Wider and narrower concepts are shown in a new "Anchoring" section of the fact inspector.
* Labels and documentation from wider concepts are added to the search index for the facts to which they relate, so a search for the name or description of a wider concept will find related narrower facts.
* Search now allows search for element names, as well as labels.

The anchoring section of the fact inspector is only shown if there is at least one anchoring relationship in the document.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/fact.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.js
@@ -253,36 +253,19 @@ Fact.prototype.isHidden = function () {
 }
 
 Fact.prototype.widerConcepts = function () {
-    if (!this._report.usesAnchoring()) {
-        return [];
-    }
     var concepts = [];
-    const wn = this._report.data.rels["w-n"];
-    for (const elr in wn) {
-        for (const src in wn[elr]) {
-            if ($.map(wn[elr][src], (target) => target.t).includes(this.conceptName())) {
-                concepts.push(src);
-            }
-        }
+    const parentsByELR = this._report.getParentRelationships(this.conceptName(), "w-n");
+    for (const elr in parentsByELR) {
+        concepts.push(...$.map(parentsByELR[elr], (rel) => rel.src));
     }
     return concepts;
 }
 
 Fact.prototype.narrowerConcepts = function () {
-    if (!this._report.usesAnchoring()) {
-        return [];
-    }
     var concepts = [];
-    const wn = this._report.data.rels["w-n"];
-    for (const elr in wn) {
-        for (const src in wn[elr]) {
-            if (src == this.conceptName()) {
-                for (const target of wn[elr][src]) {
-                    concepts.push(target.t)
-                }
-            }
-        }
+    const childrenByELR = this._report.getChildRelationships(this.conceptName(), "w-n");
+    for (const elr in childrenByELR) {
+        concepts.push(...$.map(childrenByELR[elr], (rel) => rel.t));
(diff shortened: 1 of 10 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -251,3 +251,19 @@ Fact.prototype.footnotes = function () {
...
+
+Fact.prototype.widerConcepts = function () {
+    if (!this._report.usesAnchoring()) {
+        return [];
+    }
+    var concepts = [];
+    const wn = this._report.data.rels["w-n"];
+    for (const elr in wn) {
+        for (const src in wn[elr]) {
+            if ($.map(wn[elr][src], (target) => target.t).includes(this.conceptName())) {
```
brettkail-wk, line 263: I guess this is loop is O(anchors), so it might be nice to build more direct maps (here and narrower).  Well, for this wider code, the `$.map` is generating temp arrays for every comparison.  Perhaps `$.grep(wn[elr][src], (target) => target.t.includes(concept))` (or `wn[elr][src].some(...)` if we can support new enough JavaScript?) would avoid at least avoid the temp array.

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -267,3 +267,22 @@ Fact.prototype.widerConcepts = function () {
...
+Fact.prototype.narrowerConcepts = function () {
+    if (!this._report.usesAnchoring()) {
+        return [];
+    }
+    var concepts = [];
+    const wn = this._report.data.rels["w-n"];
+    for (const elr in wn) {
+        for (const src in wn[elr]) {
+            if (src == this.conceptName()) {
+                for (const target of wn[elr][src]) {
```
brettkail-wk, line 280: Lookup the key directly rather than iterating?
```
for (const elr in wn) {
  const targets = wn[elr][this.conceptName()];
  if (targets) {
    for (const target of targets) {
```

iXBRLViewerPlugin/viewer/src/js/fact.js
brettkail-wk: I should say: I don't feel strongly about this, just discussing.

paulwarren-wk: The reverse lookup was certainly very inefficient.  We already have `getChildRelationships` for the calculations code, so it makes sense to re-use that, and introduce an equivalent `getParentRelationships` which maintains a cache of parents indexed by child.  This is done in 3a8172453f244c01872d5d7cdb249599de8f248c.  I think the first call is O(anchorsInReport), and subsequent calls are O(anchorsForConcept), which seems reasonable.

iXBRLViewerPlugin/viewer/src/js/fact.js
paulwarren-wk: Yeah, not sure what I was thinking there!   Refactored in 3a8172453f244c01872d5d7cdb249599de8f248c (see previous comment).

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully.

@Workiva/release-management-p

PR #118, opened 2020-07-07
Missing concept definitions from relationship sources

### LINKED ISSUE #116: "Building search index" is hanging when w-n present
When w-n is present, the search index gets an error: 
TypeError: Cannot read property 'labels' of undefined

How can I test the wider/narrow option? 

I can share the taxonomy package I'm using, but it is more than 10MB, so cannot upload it. Let me know which informations I should provide to make it possible to debug.

paulwarren-wk:
This PR fixes #116 where concepts that not used in the report, but which are the source of a relationship do not get added to the concept information in the report.

In the case of wider-narrower relationships where the source (wider) concept is not used in the report, this causes an exception when building the search index.

For other relationships (calculations) where the source concept is not used in the report, the viewer will load as they are not added to the search index, but an exception would occur if and when the viewer attempts to get a label for them.

As well as fixing the failure to include the source information, this PR also makes the viewer cope more gracefully with an attempt to get a label for a non-existent concept, by returning a placeholder and logging a message on the console.

### DIFF
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -178,6 +178,7 @@ class IXBRLViewerBuilder:
                         rel['w'] = r.weight
                     rr.setdefault(fromKey, []).append(rel)
                     self.addConcept(r.toModelObject)
+                    self.addConcept(r.fromModelObject)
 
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully. Smoke tested successfully.

@Workiva/release-management-p

PR #129, opened 2020-08-24
Handle nil text facts

paulwarren-wk:
Previously, nil-valued text facts would cause a JavaScript error when clicking on them.  This PR fixes this, and the value should be displayed as "nil" in the fact inspector.

Additionally, this PR applies italics consistently to "nil" (value) and "n/a" (units and accuracy) when they appear in the fact inspector, making a nil value distinct from a string value of `nil`.

Fixes XT 790

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/fact.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.js
@@ -92,6 +92,9 @@ Fact.prototype.readableValue = function() {
             v = formattedNumber + " " + this.unit().valueLabel();
         }
     }
+    else if (this.isNil()) {
+        v = "nil";
+    }
     else if (this.escaped()) {
         var html = $("<div>").append($($.parseHTML(v, null, false)));
         /* Insert an extra space at the beginning and end of block elements to
--- a/iXBRLViewerPlugin/viewer/src/js/inspector.js
+++ b/iXBRLViewerPlugin/viewer/src/js/inspector.js
@@ -474,9 +474,9 @@ Inspector.prototype._updateValue = function (item, showAll, context) {
         $('tr.value', context).removeClass('truncated');
     }
 
-    $('tr.value td .value', context).empty().text(v);
+    var valueSpan = $('tr.value td .value', context).empty().text(v);
     if (item instanceof Fact && item.isNil()) {
-        $('tr.value td .value', context).wrapInner("<i></i>");
+        valueSpan.wrapInner("<i></i>");
     }
 
 }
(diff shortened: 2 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -92,6 +92,9 @@ Fact.prototype.readableValue = function() {
             v = formattedNumber + " " + this.unit().valueLabel();
         }
     }
+    else if (this.isNil()) {
+        v = "nil";
```
brettkail-wk, line 96: Move this above `isNumeric` so the `isNil` handling can be common?

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -473,7 +474,11 @@ Inspector.prototype._updateValue = function (text, showAll, context) {
         $('tr.value', context).removeClass('truncated');
     }
 
-    $('tr.value td .value', context).text(v);
+    $('tr.value td .value', context).empty().text(v);
+    if (item instanceof Fact && item.isNil()) {
+        $('tr.value td .value', context).wrapInner("<i></i>");
```
brettkail-wk, line 479: You could capture the jq object to avoid the double lookup for text+wrapInner (here and `_selectionSummaryAccordian`).

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -531,8 +536,13 @@ Inspector.prototype._selectionSummaryAccordian = function() {
                 );
             }
             this._updateEntityIdentifier(fact, factHTML);
-            this._updateValue(fact.readableValue(), false, factHTML);
-            $('tr.accuracy td', factHTML).text(fact.readableAccuracy());
+            this._updateValue(fact, false, factHTML);
+
+            $('tr.accuracy td', factHTML).empty().append(fact.readableAccuracy());
```
brettkail-wk, line 541: Why change from `.text` to `.append`?  Why `.append` here but `.text` in `_updateValue`?

paulwarren-wk: Good point.  I briefly toyed with having readableAccuracy return jQuery objects, so switched to append.  Will change it back.

iXBRLViewerPlugin/viewer/src/js/fact.js
paulwarren-wk: It's not quite that simple, as for numerics we still include units in the resulting string, and "nil" just replaces the numeric part, so we'd still need a second if.

Whether it makes sense to include units alongside "nil" is another question, but it's information that's in the report so I figure we should display it.

iXBRLViewerPlugin/viewer/src/js/inspector.js
paulwarren-wk: Fixed in 4009081

iXBRLViewerPlugin/viewer/src/js/inspector.js
paulwarren-wk: Fixed in 4009081.

iXBRLViewerPlugin/viewer/src/js/fact.js
brettkail-wk: Ah! I overlooked that.  Looks fine to me, thanks :-).

derekgengenbacher-wf:
QA +10 Tests pass successfully. Builds successfully. Ran this branch against doc that had `nil` in it. It successfully displayed it.

@Workiva/release-management-p

PR #153, opened 2020-11-05
Typed dimensions support

paulwarren-wk:
This adds support for typed dimensions in the viewer.

Typed dimensions were previously dropped silently and would not be shown in the inspector.  It would also result in facts that differed only in typed dimension values would be treated as duplicate facts.

Typed dimension values are now shown in the inspector.  nil-valued typed dimensions are shown with a value of  `<nil>`

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/aspect.js
+++ b/iXBRLViewerPlugin/viewer/src/js/aspect.js
@@ -58,13 +58,17 @@ Aspect.prototype.isTaxonomyDefined = function() {
     return (this._aspect.indexOf(":") > -1);
 }
 
+Aspect.prototype.isNil = function() {
+    return this._value === null;
+}
+
 Aspect.prototype.valueLabel = function(rolePrefix) {
     if (this._aspect == 'c') {
         return this._report.getLabel(this._value, rolePrefix) || this._value;
     }
     if (this.isTaxonomyDefined()) {
         if (this._report.getConcept(this._aspect).isTypedDimension()) {
-            return this._value === null ? "<nil>" : this._value;
+            return this._value === null ? "nil" : this._value;
         }
         return this._report.getLabel(this._value, rolePrefix) || this._value;
     }
--- a/iXBRLViewerPlugin/viewer/src/js/aspect.js
+++ b/iXBRLViewerPlugin/viewer/src/js/aspect.js
@@ -64,7 +64,7 @@ Aspect.prototype.valueLabel = function(rolePrefix) {
     }
     if (this.isTaxonomyDefined()) {
         if (this._report.getConcept(this._aspect).isTypedDimension()) {
-            return this._value === null ? "<null>" : this._value;
+            return this._value === null ? "<nil>" : this._value;
         }
         else {
             return this._report.getLabel(this._value, rolePrefix) || this._value;
--- a/iXBRLViewerPlugin/viewer/src/js/aspect.js
+++ b/iXBRLViewerPlugin/viewer/src/js/aspect.js
@@ -53,10 +53,22 @@ Aspect.prototype.equalTo = function(a) {
     return a !== undefined && this._aspect == a._aspect && this._value == a._value;
 }
 
+
+Aspect.prototype.isTaxonomyDefined = function() {
+    return (this._aspect.indexOf(":") > -1);
+}
+
 Aspect.prototype.valueLabel = function(rolePrefix) {
-    /* Taxonomy-defined dimension, treat as explicit - or concept */
-    if (this._aspect.indexOf(":") > -1 || this._aspect == 'c') {
+    if (this._aspect == 'c') {
         return this._report.getLabel(this._value, rolePrefix) || this._value;
+    }
+    if (this.isTaxonomyDefined()) {
+        if (this._report.getConcept(this._aspect).isTypedDimension()) {
+            return this._value === null ? "<null>" : this._value;
+        }
+        else {
+            return this._report.getLabel(this._value, rolePrefix) || this._value;
+        }
     }
     else if (this._aspect == 'u') {
         if (this._value === null) {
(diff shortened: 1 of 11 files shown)


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/search.js
```diff
@@ -35,7 +35,14 @@ ReportSearch.prototype.buildSearchIndex = function () {
         doc.startDate = f.periodFrom();
         var dims = f.dimensions();
         for (var d in dims) {
-            l += " " + this._report.getLabel(dims[d],"std");
+            if (this._report.getConcept(d).isTypedDimension) {
+                if (dims[d] !== null) {
+                    l += " " + d;
```
brettkail-wk, line 40: Why dimension concept rather than member for typed?  Alternatively, why not dimension concept also for explicit?  If keeping dimension concept, why exclude if nil?

iXBRLViewerPlugin/viewer/src/js/aspect.js
```diff
@@ -53,11 +53,23 @@ Aspect.prototype.equalTo = function(a) {
...
-    /* Taxonomy-defined dimension, treat as explicit - or concept */
-    if (this._aspect.indexOf(":") > -1 || this._aspect == 'c') {
+    if (this._aspect == 'c') {
         return this._report.getLabel(this._value, rolePrefix) || this._value;
     }
+    if (this.isTaxonomyDefined()) {
+        if (this._report.getConcept(this._aspect).isTypedDimension()) {
+            return this._value === null ? "<nil>" : this._value;
+        }
+        else {
```
sagesmith-wf, line 69: Nit: else unneeded because of the early return

sagesmith-wf (approved):

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/js/search.js
paulwarren-wk: This was a typo - it should be the dimension value.

iXBRLViewerPlugin/viewer/src/js/aspect.js
paulwarren-wk: Fixed in 0b923e0

iXBRLViewerPlugin/viewer/src/js/aspect.js
```diff
@@ -53,11 +53,23 @@ Aspect.prototype.equalTo = function(a) {
...
+
 Aspect.prototype.valueLabel = function(rolePrefix) {
-    /* Taxonomy-defined dimension, treat as explicit - or concept */
-    if (this._aspect.indexOf(":") > -1 || this._aspect == 'c') {
+    if (this._aspect == 'c') {
         return this._report.getLabel(this._value, rolePrefix) || this._value;
     }
+    if (this.isTaxonomyDefined()) {
+        if (this._report.getConcept(this._aspect).isTypedDimension()) {
+            return this._value === null ? "<nil>" : this._value;
```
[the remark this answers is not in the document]
paulwarren-wk: It was a lazy attempt to differentiate it from a typed dimension with a string value of "nil".  I've now done it properly and consistently (615354f) and used italics.

I've also removed nil dimension values from search results (and the list of linked facts for a footnote).

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -139,16 +139,16 @@ Fact.prototype.isMonetaryValue = function () {
 }
 
 Fact.prototype.aspects = function () {
-    var aspects = {};
-    var fact = this;
-    $.each(this.f.a, function (k,v) {
-        aspects[k] = fact.aspect(k);
-    });
+    var aspects = [];
+    $.each(this.f.a, (k,v) => { aspects.push(this.aspect(k) ) } );
```
brettkail-wk, line 143: This is of course fine as-is, but `return $.map(this.f.a, (v, k) => this.aspect(k))` might be simpler.

brettkail-wk (approved):

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/fact.js
paulwarren-wk: Good call, although in the name of dropping unnecessary JQuery, I've gone with `return Object.keys(this.f.a).map(k => this.aspect(k));`

brettkail-wk: Ah, I wasn't sure if we could depend on those, but caniuse.com suggests they're fine.  With the `v` param now gone, I guess `.map(this.aspect)` would further simplify (obviously fine as-is :-)).

derekgengenbacher-wf (approved):

iXBRLViewerPlugin/viewer/src/js/fact.js
paulwarren-wk: Turns out that doesn't work because it's an object method, rather than a straight function.  Apparently you can make it work with `.map(this.aspect.bind(this))` but I think that's a step backwards.

Re `.map` - I'm not sure if we have an official line on browser support for the viewer.  We originally agreed to target Chrome, but it mostly works on FireFox.  I've started using `for` ... `of` in various places, which doesn't work on IE (fine on Edge).

brettkail-wk: Argh, I forgot `this` binding :-(.  I agree `.bind(this)` is worse.  Sorry for the distraction.

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully. Built locally and created viewer from provided docs. Verified viewer now shows typed dimensions correctly.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #157, opened 2020-11-09
Multi fact in cell selection fix

paulwarren-wk:
This fixes a bug whereby if the contents of a table cell are double-tagged (i.e. two or more nested tags) then clicking on the table cell would only show a single fact in the fact inspector.  

Clicking on the cell now results in both facts being shown as an accordian, as we do for other nested facts.

The bug is triggered by the special handling applied when a tagged value is the only significant content in a table cell.  In this case, we treat the table cell as the "wrapper" element that we apply highlighting/borders to as this looks much better than applying it to just the value.  In this scenario, multi-tagged content means we only have a single element to attach the fact ID to.  This fix changes the "ivid" fact data to an array, rather than a single value.


aviary2-wf:
## Security Insights

## Action Items
---

brettkail-wk (approved):

derekgengenbacher-wf (approved):

sagesmith-wf (approved):

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully. Built locally and created ixbrl viewer from testing docs attached to ticket. Verified successfully that a cell with multiple tagged facts appears correctly in the inspector as an accordion.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #206, opened 2021-03-11
XT-1255 Format fact with commas and decimals when accuracy is inf

sagesmith-wf:
### Description
No commas were being added when a fact value was set to infinite precision. This adjusts the code to format inf accuracy facts

| Current | Previous |
|---------|----------|
| Screen Shot 2021-03-11 at 2 29 03 PM |Screen Shot 2021-03-11 at 2 29 16 PM  |

### Testing Steps
1. Create a doc with a fact that has a value large enough to have commas inserted
2. Set the fact accuracy to INF
3. Generate
4. Open the preview and verify that there are no commas in the fact value
5. Deploy this build
6. Regenerate
7. Open the preview and verify that there are now commas in the fact value

### Review
@Workiva/xt

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/fact.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.js
@@ -16,7 +16,7 @@ import { isodateToHuman } from "./util.js"
 import { QName } from "./qname.js"
 import { Aspect } from "./aspect.js";
 import { Period } from './period.js';
-import { formatNumber } from "./util.js";
+import { formatNumber, numberWithCommas } from "./util.js";
 import { Footnote } from "./footnote.js";
 import $ from 'jquery'
 
@@ -77,7 +77,7 @@ Fact.prototype.readableValue = function() {
             formattedNumber = "nil";
         }
         else if (d === undefined) {
-            formattedNumber = formatNumber(v,0);
+            formattedNumber = numberWithCommas(v);
         }
         else {
             if (d < 0) {
--- a/iXBRLViewerPlugin/viewer/src/js/fact.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.js
@@ -77,7 +77,7 @@ Fact.prototype.readableValue = function() {
             formattedNumber = "nil";
         }
         else if (d === undefined) {
-            formattedNumber = v;
+            formattedNumber = formatNumber(v,0);
         }
         else {
             if (d < 0) {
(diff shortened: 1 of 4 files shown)


austinmatherne-wk (approved):

aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -77,7 +77,7 @@ Fact.prototype.readableValue = function() {
             formattedNumber = "nil";
         }
         else if (d === undefined) {
-            formattedNumber = v;
+            formattedNumber = formatNumber(v,0);
```
brettkail-wk, line 80: Add a test [like this](https://github.com/Workiva/ixbrl-viewer/blob/6e57b52f426560a29c2d4a586cecb03f2c56a5d8/iXBRLViewerPlugin/viewer/src/js/fact.test.js#L122-L139) but with a large integral part [like this](https://github.com/Workiva/ixbrl-viewer/blob/6e57b52f426560a29c2d4a586cecb03f2c56a5d8/iXBRLViewerPlugin/viewer/src/js/fact.test.js#L102-L120).

As-is, it looks like this change incorrectly formats INF `1234.56` as `1,234` rather than `1,234.56`.

derekgengenbacher-wf (approved):

iXBRLViewerPlugin/viewer/src/js/fact.js
sagesmith-wf: Thanks Brett :). First ever javascript experience, fumbling my way through this.

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -35,6 +35,13 @@ export function momentToHuman(d, adjust) {
     return d.format("D MMM Y HH:mm:ss");
 }
 
+/*
+ * Format a number with a thousands separator, infinite decimal places
+ */
+export function numberWithCommas(v) {
+    return Number(v).toString().replace(/\B(?<!\.\d*)(?=(\d{3})+(?!\d))/g, ",");
```
brettkail-wk, line 42: Why is this regex different from `formatNumber`?  Perhaps just change `formatNumber` to this:
```
    var n = Number(v);
    var s = d === undefined ? n.toString() : n.toFixed(Math.max(0, d));
    return d.replace(...);
```
...and then remove all the special-case handling from `readableValue`?

brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Builds successfully. Built locally and tested against doc with 10000000000. It was correctly formatted with commas.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #201, opened 2021-03-02
Enable localization of the viewer through i18next library

huizarmx:
I think the iXBRL viewer should have support for localization, that's why I would like to propose the following change to have the user interface in spanish or english as needed. It's also possible to add more languages maybe through a small refactor to put in separate files the translation resources.

Please let me know if this is of interest to you and I would gladly work on changes to better implement this.

Kindest Regards,

Issue #203, opened 2021-03-03
Values other than numeric are not exported to the excel document


huizarmx:
When the table to export contains data of a data type that it isn't numeric, it just display NaN in the resulting excel file.

It would be nice to get those facts exported in the excel as well

PR #218, opened 2021-04-21
Internationalization support

paulwarren-wk:
This add support for internationalization of UI components to the viewer.  It is based on https://github.com/Workiva/ixbrl-viewer/pull/201 , but puts the translations into separate JSON files, and switches languages based on the browser's detected language.

The browser's detected language can be overridden by specifying a query parameter, e.g.  `?lang=es`


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/aspect.js
```diff
@@ -74,19 +75,11 @@ Aspect.prototype.valueLabel = function(rolePrefix) {
...
-            if (qname.localname == 'GBP') {
-                return "£";
-            }
-            else if (qname.localname == 'USD') {
-                return "US $";
-            }
-            else if (qname.localname == 'EUR') {
-                return "€";
-            }
+            return i18next.t(`currencies.unitFormat${qname.localname}`, {defaultValue: qname.localname + ' '});
```
brettkail-wk, line 82: This change seems odd.  Do different countries actually display the currency symbol differently?

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -299,7 +324,7 @@ Inspector.prototype._anchorList = function (fact, anchors) {
         }
     }
     else {
-        $("<li><i>None</i></li>").appendTo(html);
+        $(`<li><i>${i18next.t("common.none")}</i></li>`).appendTo(html);
```
brettkail-wk, line 327: It seems like we should avoid interpolation of any kind into HTML strings.  Can we either use `.text` (on the inner `<i>`) or use `data-i18n` and `.localize()`?

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -19,9 +19,9 @@ import { Viewer } from "./viewer.js";
 import { Inspector } from "./inspector.js";
 
 export function iXBRLViewer() {
-    this._plugins = [];
-    this.inspector = new Inspector(this);
-    this.viewer = null;
+	this._plugins = [];
+	this.inspector = new Inspector(this);
+	this.viewer = null;
```
brettkail-wk, line 24: The rest of the repo uses spaces, so this change to tabs seems wrong.

iXBRLViewerPlugin/viewer/src/js/tableExport.js
```diff
@@ -159,17 +159,27 @@ TableExport.prototype._writeTable = function (data) {
...
-                cc.numFmt = '#,##0';
-                ws.getColumn(j+1).width = 18;
-                /* Make this an option - apply presentation signs */
-                if (cell.negative) {
-                    cc.value = Math.abs(cc.value) * -1;
+                var numValue = Number(cell.fact.value());
+                if (Number.isNaN(numValue)) {
+                    cc.value = cell.fact.value();
+                    if (/<\/?[a-z][\s\S]*>/i.test(cc.value))  {
+                        cc.value = strInputCode.replace(/<\/?[^>]+(>|$)/g, "");
```
brettkail-wk, line 166: What is the intent of this new `isNaN` branch, and how does it related to i18n?  These seem like they're trying to heuristically match `<` markup of some kind, but I don't know why.

iXBRLViewerPlugin/viewer/src/js/aspect.js
paulwarren-wk: I'm not sure (@huizarmx may know otherwise?).  It's possible that some locales may use a symbol where we'd prefer to use the ISO code because the symbol may unfamiliar.

In any event, I think this is a reasonable approach as we've moved a 3-way if block into a single line lookup in a separate JSON file, which is already being used for translating other parts of monetary/numeric amounts (e.g. translating "cents" or "millions").

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
paulwarren-wk: Fixed in a0ef97d

iXBRLViewerPlugin/viewer/src/js/inspector.js
paulwarren-wk: Fixed in fcaaa71 (with `.text`)

iXBRLViewerPlugin/viewer/src/js/tableExport.js
paulwarren-wk: Looks like it's fixing #203.

It's unrelated to i18n so I've moved to a new branch https://github.com/paulwarren-wk/ixbrl-viewer/tree/203-non-numeric-table-export

I'll review it and raise a separate PR.

iXBRLViewerPlugin/viewer/src/js/aspect.js
huizarmx: I've seen cases where the user would want the currency symbol on the right side. I think is a simple solution and brings flexibility for displaying different currencies.

brettkail-wk: Ok, thanks.

However, since the `unitFormat` in `es` are currently the same as `en`, does `fallbackLang` allow us to remove all of those so that we don't have to duplicate them in every translation?

huizarmx: Of course, that's very reasonable.

derekgengenbacher-wf:
@paulwarren-wk This is failing unit tests.

iXBRLViewerPlugin/viewer/src/js/aspect.js
paulwarren-wk: Now done in 6fa928b

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -191,38 +192,23 @@ Fact.prototype.isNil = function() {
 
 Fact.prototype.readableAccuracy = function () {
     if (!this.isNumeric() || this.isNil()) {
-        return "n/a";
+        return i18next.t("common.notAvailable");
```
brettkail-wk, line 195: This is obviously fine, but I would have expected `notApplicable`.

iXBRLViewerPlugin/viewer/src/html/inspector.html
```diff
@@ -31,8 +31,8 @@
   <div id="inspector" class="no-fact-selected">
     <div id="inspector-head">
       <div class="title">
-        <h1 class="footnote-mode-off">Fact Properties</h1>
-        <h1 class="footnote-mode-only">Footnote Properties</h1>
+        <h1 class="footnote-mode-off" data-i18n="inspector.factProperties">Fact Properties</h1>
+        <h1 class="footnote-mode-only" data-i18n="inspector.footNoteProperties">Footnote Properties</h1>
```
brettkail-wk, line 35: This is obviously fine, but I would have expected `footnoteProperties`.

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -701,10 +733,21 @@ Inspector.prototype.switchItem = function (id) {
     this.update();
 }
 
+Inspector.prototype.preferredLanguages = function () {
+    var urlParams = new URLSearchParams(window.location.search);
+    if (urlParams.has("lang")) {
+        return [ urlParams.get("lang") ];
+    }
+    var langs = window.navigator.languages || [ window.navigator.language || window.navigator.userLanguage ] ;
```
brettkail-wk, line 741: Can `language` and `userLanguage` both be null (null reference in the `.each` below), or does that never happen in practice for modern browsers?

brettkail-wk (approved):
Tests are still failing (as Derek mentioned).

brettkail-wk (dismissed):

iXBRLViewerPlugin/viewer/src/html/inspector.html
paulwarren-wk: Done in c975ed1

iXBRLViewerPlugin/viewer/src/js/fact.js
paulwarren-wk: Fixed in a4bf05a

iXBRLViewerPlugin/viewer/src/js/inspector.js
paulwarren-wk: I suspect it's safe, but I've made it safer in 87ed933

brettkail-wk (dismissed):

paulwarren-wk:
4167b1f fixes a bug introduced by this PR that prevented the analysis chart from opening.

brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
QA +10 Unit test pass successfully. Built locally and tested against local doc. language override works successfully.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #219, opened 2021-04-22
Fix scrolling to show selected elements on documents with scrollable divs

paulwarren-wk:
A number of ESEF filings don't have scroll bars on the `<body>` element but instead have a `<div>` element that takes up the whole viewport and has a scrollbar.  With such documents, the viewer would not scroll the viewport to show the selected fact.

This PR fixes this, by scrolling any scrollable ancestor elements as required to make an element visible.

The PR also fixes a small bug whereby inline styles and other attributes on the `<body>` tag were not copied into the inspector's iframe.

A sample document with both inline body styles and content in a scrollable div is included.  A real world example of a document with the scrollable div problem can be seen here:

https://filings.xbrl.org/5493007JXOLJ0QCY2D70/2020-12-31/ESEF/GB/

The PR also fixes scrolling to elements where the tag is not actually visible because it's hidden inside an element with `display: none` (which is the case for most of the tags in the above ESEF example!)

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
@@ -113,8 +113,10 @@ iXBRLViewer.prototype._reparentDocument = function () {
     const iframeBody = $(iframe).contents().find('body');
     $('body').children().not("script").not('#ixv').not(iframeContainer).appendTo(iframeBody);
 
-    for (const bodyAttr of $('body').prop("attributes")) {
+    /* Move all attributes on the body tag to the new body */
+    for (const bodyAttr of [...$('body').prop("attributes")]) {
         iframeBody.attr(bodyAttr.name, bodyAttr.value); 
+        $('body').removeAttr(bodyAttr.name);
     }
 
     /* Avoid any inline styles on the old body interfering with the inspector */
--- a/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
@@ -110,7 +110,12 @@ iXBRLViewer.prototype._reparentDocument = function () {
     /* Due to self-closing tags, our script tags may not be a direct child of
      * the body tag in an HTML DOM, so move them so that they are */
     $('body script').appendTo($('body'));
-    $('body').children().not("script").not('#ixv').not(iframeContainer).appendTo($(iframe).contents().find('body'));
+    const iframeBody = $(iframe).contents().find('body');
+    $('body').children().not("script").not('#ixv').not(iframeContainer).appendTo(iframeBody);
+
+    for (const bodyAttr of $('body').prop("attributes")) {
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Security Insights
<details>
<summary>(4) Security relevant changes were detected</summary>
<li>Watched keyword <code>etree</code> in <code>iXBRLViewerPlugin/iXBRLViewer.py</code> line(s) ['199', '322'] added</li>
<li>Watched keyword <code>etree</code> in <code>iXBRLViewerPlugin/iXBRLViewer.py</code> line(s) ['300'] removed</li>
<li>Watched keyword <code>etree</code> in <code>iXBRLViewerPlugin/xhtmlserialize.py</code> line(s) ['102'] added</li>
</details>

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -110,7 +110,12 @@ iXBRLViewer.prototype._reparentDocument = function () {
     /* Due to self-closing tags, our script tags may not be a direct child of
      * the body tag in an HTML DOM, so move them so that they are */
     $('body script').appendTo($('body'));
-    $('body').children().not("script").not('#ixv').not(iframeContainer).appendTo($(iframe).contents().find('body'));
+    const iframeBody = $(iframe).contents().find('body');
+    $('body').children().not("script").not('#ixv').not(iframeContainer).appendTo(iframeBody);
+
+    for (const bodyAttr of $('body').prop("attributes")) {
```
brettkail-wk, line 116: Is it correct to copy all attributes, or should we only copy a subset (style, class)?

Should we "move" rather than "copy" the attributes (we only remove `style` below)?

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
paulwarren-wk: Both good questions :-)

The Python code doesn't add or remove any attributes to the `<body>` tag, so any attributes that are there will have come from the original iXBRL, so I think "move all" is probably what we want, as just about anything can potentially be used to influence styling. For example, an `id` attribute could be referenced in CSS.  Other attributes not currently permitted in iXBRL could also be used, such as `lang` and `data-` attributes.

derekgengenbacher-wf (approved):

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
paulwarren-wk: 17fc967d55f4da5ab00c42d15eb969c427a1d35a implements moving all attributes.

derekgengenbacher-wf (approved):

brettkail-wk (approved):

sagesmith-wf (approved):

derekgengenbacher-wf:
QA +10 unit tests pass successfully. Builds successfully. Built locally and tested against provided doc successfully.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #224, opened 2021-05-02
Avoid missing elements in calculation inspector and improve label language fallback

paulwarren-wk:
Previously, if a concept that did not have an English label (or a label in the current viewer language, if different) was shown in the calculation section of the inspector, a blank line was shown.

This PR fixes this by changing the label language fallback to be:

1. Current viewer language
2. en
3. en-US
4. (new) Any defined label (chosen deterministically based on language code in alphabetical order)

If no label is found, the calculation view will now fall back on concept name, so a blank line is never displayed.

The above fallback hierarchy is now used in all places where a label is displayed.

Fixes XT-1354

PR #228, opened 2021-05-03
Extensible enumerations support

paulwarren-wk:
This PR adds support for extensible enumeration facts.

Previously, such facts showed the raw value in the fact inspector, e.g. `http://fasb.org/us-gaap/2021-01-31#AccruedLiabilitiesMember`

These facts now show the standard label for the member(s) in the enumeration value, e.g.:

image

Fixes XT-1334


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/__init__.py
```diff
@@ -40,6 +40,9 @@
 from .localviewer import launchLocalViewer, VIEWER_BASENAME_SUFFIX
 from arelle.ModelDocument import Type
 
+import traceback
+import sys
```
brettkail-wk, line 44: Remove unused?

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -219,9 +222,21 @@ def createViewer(self, scriptUrl="js/dist/ixbrlviewer.js"):
...
-                "v": f.value if not f.isNil else None,
                 "a": aspects,
             }
+
+            if f.isNil:
+                factData["v"] = None
+            elif f.concept is not None and f.concept.isEnumeration:
+                qnEnums = f.xValue
+                if not isinstance(qnEnums, list):
+                    qnEnums = (qnEnums,)
```
brettkail-wk, line 233: What does this do if the fact value is invalid?  Do we need some kind of `qnEnums is not None`?

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -100,6 +100,13 @@ Fact.prototype.readableValue = function() {
         /* Replace runs of whitespace (including nbsp) with a single space */
         v = html.text().replace(/[\u00a0\s]+/g, " ").trim();
     }
+    else if (this.isEnumeration()) {
+        var labels = [];
+        for (const qn of v.split(' ')) {
+            labels.push(this._report.getLabel(qn, 'std') || qn);
```
brettkail-wk, line 106: Should we wait on the other PR and `getLabelOrName` here?

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -100,6 +100,13 @@ Fact.prototype.readableValue = function() {
...
+        }
+        v = labels.join(', ');
```
brettkail-wk, line 108: Do you think people would prefer to see this as a bulleted list rather than a comma-separated list?  If we don't yet have enough experience with enum set, I'm fine with this for now.

brettkail-wk (approved):

sagesmith-wf (approved):

iXBRLViewerPlugin/__init__.py
paulwarren-wk: They are used (see [line 74](https://github.com/Workiva/ixbrl-viewer/pull/228/files/1751ec0fcef4eabed6e1d81bc0a25a40fb3e3cb2#diff-a65816b2086d1b8e75aabec4dd3761c8795d7693a1d186400b4718c1a894fceaR74))

I managed to trigger an exception during development which highlighted that these imports were missing.

brettkail-wk: Ah, thanks, I overlooked the commit message, sorry.

iXBRLViewerPlugin/viewer/src/js/fact.js
paulwarren-wk: Yes, good point.  I've now merged the other branch into this one and made that change, so this PR is currently showing the changes in both, but it should go back to just this once #224 is merged.

iXBRLViewerPlugin/viewer/src/js/fact.js
paulwarren-wk: Just had a quick chat with Dave Shaw on use in the US-GAAP taxonomy, and it sounds like most cases will actually be a single item, but where is more than one, a bulleted list would be clearer, so I'll make that change.

sagesmith-wf (approved):

derekgengenbacher-wf (approved):

rmconsole-wf (approved):
+1 from RM

PR #229, opened 2021-05-08
Concealed facts

paulwarren-wk:
Due to restrictions on the content of `ix:nonNumeric` elements (no HTML markup allowed within them), some ESEF filings have taken to including numeric values twice: once, untagged, with additional elements for styling, and again inside a `display: none` span, so that it can be tagged.

This is very confusing in the viewer, as it appears that the number isn't tagged, but it's not flagged as a hidden tag either.
This PR makes it obvious where this has been done by adding a "concealed tag" label on such elements.

image

There's broad agreement that this tagging approach is very poor practice, and that if such in-tag styling is required, it's better to use a normal hidden tag and the `-esef-ix-hidden` mechanism so that there is a link between the tag and the presented number.  

Note that this branch is already merged with the i18n branch, and so should only be merged after that.


iXBRLViewerPlugin/viewer/src/html/fact-details.html
```diff
@@ -15,34 +15,34 @@
 -->
```
brettkail-wk, line 15: (Arbitrary line.)

> some ESEF filings have taken to including numeric values twice: once, untagged, with additional elements for styling, and again inside a display: none span, so that it can be tagged.

This seems like a violation of [IXBRL 9.1.2](http://www.xbrl.org/specification/inlinexbrl-part1/rec-2013-11-18/inlinexbrl-part1-rec-2013-11-18.html#d1e4467).  I assume that rule exists for the same reason that motivates this PR: the facts aren't explicitly annotated, so viewer software has no good way of categorizing the facts without a CSS engine (e.g., if we wanted to shift analysis to the plugin to reduce startup cost).  Should we work with the vendor to fix?

iXBRLViewerPlugin/viewer/src/js/ixnode.js
```diff
@@ -32,6 +32,7 @@ export function IXNode(id, wrapperNode, docIndex) {
     this.docIndex = docIndex;
     this.footnote = false;
     this.id = id;
+    this.visible = true;
```
brettkail-wk, line 35: I assume "hidden" was not used to avoid concusion with ix:hiddren, but "visible" seems to imply CSS "visibility".  Perhaps `htmlHidden` would be clearer?

Perhaps the CSS class should similarly be `html-hidden-fact-tag` for consistency?

brettkail-wk (dismissed):

iXBRLViewerPlugin/viewer/src/html/fact-details.html
paulwarren-wk: I think you're right.  I thought there was something in the spec about this, but I failed to find it despite looking several times!

We've already raised this with the vendor, and with ESMA.  I think the issue behind this is an ESEF filing rule about not putting values in `ix:hidden` that are "eligible for transformation".  The ESEF rules don't define what that means, but I think there's an equivalent SEC rule which effectively says that monetary facts are eligible for transformation and thus can't go in `ix:hidden`.  Not withstanding the above rule that makes this technically invalid, using `ix:hidden` with `-esef-ix-hidden` is clearly preferable to using a repeated value with `display: none`.

I should have mentioned that #219 made it so that the viewer will scroll to where the hidden element is, but obviously nothing is highlighted.

iXBRLViewerPlugin/viewer/src/js/ixnode.js
paulwarren-wk: Good call - done in 88edf9d

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/html/fact-details.html
brettkail-wk: It's somewhat unfortunate that we're adding code that implicitly condones vendors using CSS tricks to replace a validatable ESEF/SEC errors with a non-validatable iXBRL constraints.  I would personally be inclined to do nothing at all.  If such filings are difficult to view, well, that's exactly why those constraints exist :-).

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/html/fact-details.html
derekgengenbacher-wf: @paulwarren-wk I agree with @brettkail-wk sentiment here but can be persuaded. Do we want to move forward this change given that coding for CSS tricks seems dicey?

davecasleton-wf: @derekgengenbacher-wf Let's move forward with this change given the volume of impacted filings.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Built locally and smoke tested successfully.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #230, opened 2021-05-13
Tolerate missing concepts when fetching references

paulwarren-wk:
This PR makes the viewer more tolerant of invalid iXBRL documents with missing concept definitions.

Currently, if an iXBRL report references a concept which does not exist in the taxonomy, an exception is thrown when building the search index. This fixes this.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/concept.js
+++ b/iXBRLViewerPlugin/viewer/src/js/concept.js
@@ -23,7 +23,7 @@ export function Concept(report, name) {
  * the concept has none.
  */
 Concept.prototype.referenceValuesAsString = function() {
-    if (!this._c.r) {
+    if (!this._c || !this._c.r) {
         return "";
     }
     else {
@@ -36,7 +36,7 @@ Concept.prototype.referenceValuesAsString = function() {
 }
 
 Concept.prototype.references = function () {
-    if (!this._c.r) {
+    if (!this._c || !this._c.r) {
         return  [];
     }
     else {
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/concept.js
```diff
@@ -36,7 +36,7 @@ Concept.prototype.referenceValuesAsString = function() {
 }
 
 Concept.prototype.references = function () {
-    if (!this._c.r) {
+    if (!this._c || !this._c.r) {
```
brettkail-wk, line 39: Should `isTypedDimension` be similarly adjusted?  It looks like something similar would happen [here](https://github.com/Workiva/ixbrl-viewer/blob/f69a4f82edc68403305f59644708a15c7d15e23a/iXBRLViewerPlugin/viewer/src/js/search.js#L38) if a fact has a reference to an invalid concept.  Actually, that line looks like it has a bug and should be `isTypedDimension()`?

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/concept.js
paulwarren-wk: Good point, and yes, there's a bug there which is why we weren't tripping over on undefined concepts.  Both now fixed.

sagesmith-wf (approved):

brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Built locally and tested successfully on hand modified doc.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #248, opened 2021-07-28
Improve support for invalid iXBRL values

paulwarren-wk:
This PR improves display of invalid iXBRL values.  For example, a formatted number such as `38,000.00` which does not have a transform specified, or a value which does not match the specified transform.

Such values now show "Invalid value" rather than "NaN"

image

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/inspector.js
+++ b/iXBRLViewerPlugin/viewer/src/js/inspector.js
@@ -529,7 +529,7 @@ Inspector.prototype._updateValue = function (item, showAll, context) {
     }
 
     var valueSpan = $('tr.value td .value', context).empty().text(v);
-    if (item instanceof Fact && item.isNil()) {
+    if (item instanceof Fact && (item.isNil() || item.isInvalidIXValue())) {
         valueSpan.wrapInner("<i></i>");
     }
 
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -14,7 +14,7 @@
 
 from arelle import XbrlConst
 from arelle.ModelDocument import Type
-from arelle.ModelValue import QName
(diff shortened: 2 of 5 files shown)


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -529,7 +529,7 @@ Inspector.prototype._updateValue = function (item, showAll, context) {
     }
 
     var valueSpan = $('tr.value td .value', context).empty().text(v);
-    if (item instanceof Fact && item.isNil()) {
+    if (item instanceof Fact && (item.isNil() || item.isInvalidIXValue())) {
```
sagesmith-wf, line 532: Nit: Perhaps an `isValidFact` function?

sagesmith-wf (approved):

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/inspector.js
brettkail-wk: What would that do?  If you mean it to check nil also, then nil is a valid fact value, so that wouldn't help here.  If you mean it to check for other kinds of invalid state (e.g., invalid unit), it seems wrong to add `<i>` emphasis to the value here just because something else is invalid.

sagesmith-wf: Sorry, misunderstanding

derekgengenbacher-wf:
QA +10 Unit tests pass successfully. Built locally and verified new format of invalid values.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #270, opened 2021-11-15
Document outline

paulwarren-wk:
This PR adds a document outline mode to the viewer, allowing easy navigation to different tagged sections:

image

Sections are allocated based on the longest contiguous run of facts in document order that fit into each presentation group.  A fact is deemed to fit into a group if it uses a concept from the group and, for each dimension used in the group, the fact either has a specified member, or it does not have that dimension and the group allows the default member.

Note that with the exception of identifying defaults, dimensional filtering is based entirely on information in the presentation tree, not definition relationships.


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -207,6 +207,7 @@ def createViewer(self, scriptUrl="js/dist/ixbrlviewer.js"):
         self.roleMap.getPrefix(XbrlConst.documentationLabel, "doc")
         self.roleMap.getPrefix(XbrlConst.summationItem, "calc")
         self.roleMap.getPrefix(XbrlConst.parentChild, "pres")
+        self.roleMap.getPrefix(XbrlConst.dimensionDefault, "d-d")
```
brettkail-wk, line 210: @paulwarren-wk Are the javascript changes here compatible with previously generated?  Perhaps try swapping the generated .js into an existing viewer.  @derekgengenbacher-wf This should probably be released as 1.1.0.

iXBRLViewerPlugin/viewer/src/html/inspector.html
```diff
@@ -132,6 +135,11 @@ <h3 class="collapsible-header" data-i18n="inspector.footnotes">Footnotes</h3>
                <div class="collapsible-body footnotes">
                </div>
             </div>
+            <div class="collapsible-section">
+               <h3 class="collapsible-header" data-i18n="inspector.footnotes">Fact Groups</h3>
```
brettkail-wk, line 139: `inspector.footnotes` seems wrong.

iXBRLViewerPlugin/viewer/src/js/outline.js
```diff
@@ -0,0 +1,153 @@
...
+    if (this._report.getParentRelationshipsInGroup(fact.conceptName(), "pres", elr).length == 0) {
+        return false;
+    }
+    const fd = fact.dimensions();
+    const dm = this.dimensionMap[elr];
+    // Check all dimensions specified in this ELR
+    for (const [dim, spec] of Object.entries(dm)) {
+        // If a fact has a dimension, it must be in the list of permitted
+        // members, otherwise, the default member must be allowed
+        if ((dim in fd) ? !(fd[dim] in spec.members) : !spec.allowDefault) {
```
brettkail-wk, line 80: What does this do for typed dimensions?

iXBRLViewerPlugin/viewer/src/js/outline.js
```diff
@@ -0,0 +1,153 @@
...
+    var runStart = {};
+    var longestRun = {};
+    var longestRunStart = {};
+    this.dimensionMap = this._buildDimensionMap();
+    const elrs = report.relationshipGroups("pres");
+    for (const f of facts) {
+        if (f.isHidden()) {
+            continue;
+        }
+        for (const elr of elrs) {
```
brettkail-wk, line 34: I wonder how this fact/elr/dimension looping performs for reports with tends of thousands of facts.  @derekgengenbacher-wf Can you be sure to test this against large doc?

iXBRLViewerPlugin/viewer/src/js/outline.js
```diff
@@ -0,0 +1,153 @@
...
+        dimension = conceptName;
+        dimensionMap[dimension] = { members: {}, allowDefault: false};
+    }
+    for (var rel of children[elr]) {
+        if (dimension) {
+            if (this._report.dimensionDefault(dimension) == rel.t) {
+                dimensionMap[dimension].allowDefault = true;
+            }
+            else {
+                dimensionMap[dimension].members[rel.t] = true;
```
brettkail-wk, line 126: I'm a bit lost in the XBRL-vs-JS data structures here, but I see `"pres"`.  Is this ultimately trying to build the list of dimension members based on the presentation rather than definition linkbase?

paulwarren-wk: Yes - the only "real" dimensional information used is the `dimension-default` relationship.

The problem with using the definition relationships is that you have to make the assumption that there is correspondence between the presentation and dimensional ELRs, as opposed to the assumption that we make here, which is that the presentation mirrors the dimensional structure.

Doing it based on the presentation is consistent with what the [EFM](https://www.sec.gov/info/edgar/specifications/edgarfm-vol2-v59.pdf) describes for the EDGAR renderer, e.g. (6.24.2)

> If a context segment has an axis with a member for which the _axis does not have that member as a descendant in the presentation group_ does not appear in the base set, then the context will not be selected (it will be filtered out).

iXBRLViewerPlugin/viewer/src/js/outline.js
paulwarren-wk: Good question.  If a typed dimension is mentioned in the presentation then a fact would be required to have that dimension, with a value that corresponds to the QName of a concept that is a child of the typed dimension in the presentation.  The generation code doesn't cope with QName valued typed dimensions, so it would have to be a string that happened to correspond to the QName used internally by the viewer.

I think the current behaviour is harmless, but it's untested and we should probably handle typed dimensions explicitly.

I'm not sure what's normally done with typed dimensions in the presentation.  I assume that they're included with no children, if they're included at all, so we should probably make it so that if it's in the presentation, facts are required to have it.

iXBRLViewerPlugin/iXBRLViewer.py
paulwarren-wk: Yes, although I'm not sure if we should handle it as we currently do, which is to display an "outline unavailable" message:

image

Or to hide the outline icon completely if there are no presentation relationships.

iXBRLViewerPlugin/viewer/src/js/outline.js
brettkail-wk: I agree the EFM requires it, but what about other filing authorities?  It seems somewhat fragile to rely on that.

Anyway, I guess this is all just heuristics, so if you/others are comfortable with it, then ok.  Perhaps just add a comment to this code that we're relying on the presentation to match the definition, and if we find some set of filings that doesn't conform to that, we could in theory make this more robust by using the definition.

iXBRLViewerPlugin/viewer/src/js/outline.js
brettkail-wk: > I'm not sure what's normally done with typed dimensions in the presentation.  I assume that they're included with no children

I also don't know, but I sent a message to an internal channel to see if there's a convention, but my guess is the same as yours (I would be quite surprised if they're excluded altogether).

iXBRLViewerPlugin/iXBRLViewer.py
brettkail-wk: I don't feel strongly, that looks fine to me.  So long as it doesn't completely break, I think it's ok.  We have only recently advertised support for updating the JS in place for old reports, so I don't know how many users are trying to do that.  If someone is and they don't like it, we can always make a change later once we have a real use case.

iXBRLViewerPlugin/viewer/src/js/outline.js
brettkail-wk: I confirmed that's the convention.  As an example, the [Workiva 2020 10-K](https://www.sec.gov/ix?doc=/Archives/edgar/data/0001445305/000144530521000035/wk-20201231.htm) uses RevenueRemainingPerformanceObligationExpectedTimingOfSatisfactionStartDateAxis.

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/outline.js
paulwarren-wk: Thanks - implemented in 261a1af and confirmed that it appears to work with the above document.

iXBRLViewerPlugin/viewer/src/html/inspector.html
```diff
@@ -45,12 +45,12 @@ <h1 class="footnote-mode-only" data-i18n="inspector.footnoteProperties">Footnote
         </div>
       </div>
       <div class="controls">
-        <div class="outline-button">
+        <div class="outline-button" title="Document outline">
```
brettkail-wk, line 48: Out of curiosity, is it somehow possible to do i18n for `title`?

brettkail-wk (approved):

brettkail-wk (dismissed):

iXBRLViewerPlugin/viewer/src/html/inspector.html
paulwarren-wk: I did look into this at the time, but obviously didn't look hard enough as it turns out it's very easy.   Done in a4d4a35.

I'm looking at some tools (i18next-parser) to help make the translation files maintainable, but will tackle that in a separate PR.

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/outline.js
paulwarren-wk: ESEF appear to mandate the same structure as SEC.

The UK taxonomies don't, but equally, there's no correspondence between presentation and definition ELRs there, and the document outline is imperfect on non-extension taxonomies anyway, as it'll list irrelevant sections that happen to include a fact that is reported in a different section.

I've added a comment to the code.

iXBRLViewerPlugin/iXBRLViewer.py
paulwarren-wk: I discussed this with @davecasleton-wf last week, and we also had no strong feelings, so I agree that we should leave it as is.

derekgengenbacher-wf (approved):

iXBRLViewerPlugin/viewer/src/js/outline.js
```diff
@@ -0,0 +1,168 @@
...
+        else if ((dim in fd) ? !(fd[dim] in spec.members) : !spec.allowDefault) {
+            return false;
+        }
+    }
+    return true;
+}
+
+// Build a map of ELRs to dimensional information:
+//   { elr: 
+//      { dimensionQName:
```
sagesmith-wf, line 94: small nit: probably a bit easier with the keys all on a newline and the brackets on the same line as the key thats holding the map i.e. something like:

```
{ 
    elr: {
         dimensionQname: {
                allowDefaults: true,
                         members: {
                                   memberQname: true
                         }
           }
}
```

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/js/outline.js
derekgengenbacher-wf: I tested this and it definitely slows the loading of the js on the order of about 10 seconds.

derekgengenbacher-wf:
QA +10 Unit and Skynet tests pass successfully. Builds successfully. Built js locally and tested code locally on NNN doc. Loading takes a little longer, but outline feature works correctly. Can see an issue in the future where an outline does not appropriately display and there are no indications of what could be wrong with the doc.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

iXBRLViewerPlugin/viewer/src/js/outline.js
brettkail-wk: How much slower is that?  Do we need to change it to only do it lazily or find some other way to speed it up?

derekgengenbacher-wf: Having run it a few more times in both wk-dev and locally, the time is negligible. The first few times I ran I had an outlier that took a few seconds longer than the rest of the attempts. I was also running a validation in the back ground during that run.

PR #278, opened 2022-01-20
fact-explanatoryFact support

paulwarren-wk:
This PR adds support for fact-explanatoryFact relationships.  These are footnote relationships between a pair of facts, rather than a fact and a footnote.  Prior to this PR, the viewer would crash when opening documents with these relationships.

Facts that are the target of footnote relationships are shown in the "Footnotes" section, alongside any text footnotes.

Facts that are the source of an inbound footnote relationship to another fact are shown in the "Associated facts" section, as they are currently in the footnote inspector.  The "associated facts" section is only shown for facts if there is at least one such inbound relationship.

This PR also fixes the highlighting of "related facts" (highlighting of other facts of in a calculation or search results) which previously only highlighted the main fact in an iXBRL document, and any continuation sections.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/report.js
+++ b/iXBRLViewerPlugin/viewer/src/js/report.js
@@ -44,12 +44,19 @@ iXBRLReport.prototype._initialize = function () {
     var fnorder = Object.keys(this._ixNodeMap).filter((id) => this._ixNodeMap[id].footnote);
     fnorder.sort((a,b) => this._ixNodeMap[a].docOrderindex - this._ixNodeMap[b].docOrderindex);
 
-    // Create footnote objects for all footnotes, and associate facts with
-    // those footnotes to allow 2 way fact <-> footnote navigation.
-    for (var id in this.data.facts) {
-        var f = new Fact(this, id);
-        this._items[id] = f;
-        var fns = this.data.facts[id].fn || [];
+    // Create Fact objects for all facts.  
+    for (const id in this.data.facts) {
+        this._items[id] = new Fact(this, id);
+    }
+
+    // Now resolve footnote references, creating footnote objects for "normal"
+    // footnotes, and finding Fact objects for fact->fact footnotes.  
+    //
+    // Associate source facts with target footnote/facts to allow two way
+    // navigation.
+    for (const id in this.data.facts) {
+        const f = this._items[id];
+        const fns = this.data.facts[id].fn || [];
         fns.forEach((fnid) => {
             var fn = this._items[fnid];
             if (fn === undefined) {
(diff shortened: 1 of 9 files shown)


aviary-wf:
## Security Insights
<details>
</details>

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/report.js
```diff
@@ -44,20 +44,27 @@ iXBRLReport.prototype._initialize = function () {
...
+    // Now resolve footnote references, creating footnote objects for "normal"
+    // footnotes, and finding Fact objects for fact->fact footnotes.  
+    //
+    // Associate source facts with target footnote/facts to allow two way
+    // navigation.
+    for (const id in this.data.facts) {
+        const f = this._items[id];
+        const fns = this.data.facts[id].fn || [];
         fns.forEach((fnid) => {
             var fn = this._items[fnid];
```
brettkail-wk, line 61: I'm not so familiar with the Arelle internals, so can you give a little more explanation for what's happening here?  Is Arelle somewhere processing `fact-explanatoryFact` and adding the model object to the (fake?) `XBRL-footnotes` relationship set, and then iXBRLViewer.py is copying that into `fn`, and then this line is producing either a `Footnote` or `Fact` object?  I grep'ed through Arelle looking for what would cause Arelle to do that, but I wasn't able to find anything...

paulwarren-wk: `XBRL-footnotes` appears to be a pseudo arcrole that gets populated with all XLink relationships in the instance document.

This is done in [ModelDocument.py](https://github.com/Arelle/Arelle/blob/master/arelle/ModelDocument.py#L1101):

```python
     if inInstance:
                            #index footnote links even if no arc children
                            baseSetKeys = (("XBRL-footnotes",None,None,None), 
                                           ("XBRL-footnotes",linkrole,None,None))
                            for baseSetKey in baseSetKeys:
                                self.modelXbrl.baseSets[baseSetKey].append(lbElement)
```

iXBRLViewer.py then processes all of these relationship and adds the `id` of the target to the source fact's `fn` property in the JSON data.  The `id` will either identify a footnote resource, in which case it will exist in the iXBRL document as an element with that ID, or it will be the ID of fact.

This code then checks all the footnotes on a fact, and checks if the report contains a `Fact` or `Footnote` with that ID.  If not, it assumes that it must be a text footnote, and creates a new `Footnote` object for it, and adds it to the report.

We don't add footnotes to the JSON data explicitly, as we don't need to know anything about them that we can't trivially extract from the iXBRL using JS.

paulwarren-wk:
I've just removed the "Code tidying" commits that were unrelated to this PR.

brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
QA +10 Unit and Skynet tests pass successfully. Builds successfully. Built locally and regression tested successfully.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #281, opened 2022-02-01
Improve behaviour of next/prev tag with double tagged items

paulwarren-wk:
The PR fixes and improves behaviour of the next/prev tag buttons in the presence of double-tagged facts.

Firstly, if you double tag a value which is the only content of a table cell, next/prev fact navigation would always skip over the second fact. This is now fixed, so next/prev will take you through each tag in turn.

Secondly, whenever you have double-tagged facts, clicking on the fact would show both in the inspector, but select the second (innermost) fact, meaning that "next" would take you onto a completely different fact.  This is now fixed so that it is the first fact that is selected, and "next" will take you to the second fact.   This logic is only applied for double tags (i.e. two tags with exactly the same content) not for other nested facts.

Fixes XT-1383

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -389,11 +389,13 @@ Viewer.prototype.selectElement = function (itemId, itemIdList) {
 
 // Handle a mouse click to select.  This finds all tagged elements that the
 // mouse click is within, and returns a list of item IDs for the items that
-// they're tagging.
-// The selected element is the highest ancestor which has the same content as
-// the clicked element.  This is so that when we have double tagged elements,
-// we select the first of the set, but where we have nested elements, we select
-// the innermost.
+// they're tagging.  This is so the inspector can show all items that were
+// under the click.
+// The initially selected element is the highest ancestor which is tagging
+// exactly the same content as the clicked element.  This is so that when we
+// have double tagged elements, we select the first of the set, but where we
+// have nested elements, we select the innermost, as this gives the most
+// intuitive behaviour when clicking "next".
 Viewer.prototype.selectElementByClick = function (e) {
     var itemIDList = [];
     var viewer = this;
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -234,26 +234,44 @@ Viewer.prototype.contents = function() {
     return this._iframes.contents();
 }
 
+// Move by offset (+1 or -1) through the tags in the document in document
+// order.
+//
+// Each element may have one or more tags associated with it, so we need to
+// move through the list of tags associated with the current element before
+// moving to the next/prev element
+//
+Viewer.prototype._selectAdjacentTag = function (offset, currentFact) {
+    const elements = $(".ixbrl-element:not(.ixbrl-continuation)", this._contents);
+    const currentElement = $(".ixbrl-selected:not(.ixbrl-continuation)", this._contents);
+    var nextId;
 
-Viewer.prototype._selectAdjacentTag = function (offset) {
-    var elements = $(".ixbrl-element:not(.ixbrl-continuation)", this._contents);
-    var current = $(".ixbrl-selected:not(.ixbrl-continuation)", this._contents);
-    var next;
-    if (current.length == 1) {
-        next = elements.eq((elements.index(current.first()) + offset) % elements.length);
+    if (currentElement.length == 1 && currentFact !== undefined) {
+        // The next ID may be on the current element in the case of double-tagged table cells
+        // Build a window of item IDs consisting of all IDs for the previous,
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights
<details>
</details>

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -234,26 +234,44 @@ Viewer.prototype.contents = function() {
...
-    if (current.length == 1) {
-        next = elements.eq((elements.index(current.first()) + offset) % elements.length);
+// Move by offset (+1 or -1) through the tags in the document in document
+// order.
+//
+// Each element may have one or more tags associated with it, so we need to
+// move through the list of tags associated with the current element before
+// moving to the next/prev element
+//
+Viewer.prototype._selectAdjacentTag = function (offset, currentFact) {
```
brettkail-wk, line 244: To be sure, the param is named `currentFact` (throughout), but I guess it's really `currentItem` since the inspector's current item could be a footnote, right?

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -234,26 +234,44 @@ Viewer.prototype.contents = function() {
...
+    const elements = $(".ixbrl-element:not(.ixbrl-continuation)", this._contents);
+    const currentElement = $(".ixbrl-selected:not(.ixbrl-continuation)", this._contents);
+    var nextId;
+
+    if (currentElement.length == 1 && currentFact !== undefined) {
```
brettkail-wk, line 249: Out of curiosity, what is the relevance of the `currentElement.length == 1` check?  It seems like the logic would just be "if there's a selected item, move to the prev/next selected item", but perhaps I'm overlooking some case.

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -234,26 +234,44 @@ Viewer.prototype.contents = function() {
...
+        // The next ID may be on the current element in the case of double-tagged table cells
+        // Build a window of item IDs consisting of all IDs for the previous,
+        // current and next elements, find the position of the current item ID,
+        // and move by "offset"
+
+        const prev = elements.eq((elements.index(currentElement.first()) - 1) % elements.length);
+        const next = elements.eq((elements.index(currentElement.first()) + 1) % elements.length);
+        const idWindow = prev.data('ivid').concat(currentElement.first().data('ivid'), next.data('ivid'));
+        nextId = idWindow[idWindow.indexOf(currentFact.id) + offset];
```
brettkail-wk, line 258: This `currentElement.first()` logic is quite complex, and it took me quite awhile to understand.  If I understand correctly, if we start with `<tx ivid="f1 f2"/><td ivid="f3 f4"/><td ivid="f5 f6/>` and have the middle cell selected, then we want to build an idWindow=[f1,f2,f3,f4,f5,f6] so that we can advance properly.  Is that right?

If so, I wonder if it would be simpler to just populate an `_allIds` list during `_preProcessiXBRL` and remove the `prev`/`next`/`idWindow` complication here and just use that list directly.

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: Yes, that's correct.  

We could improve the readability a bit by renaming `currentElement` to `currentElements`, and then doing `const currentElement = currentElements.first();` but a static list of IDs for the whole document might be cleaner, although it'd need to be one list per document for document sets - assuming we want to keep the current behaviour of next/prev actions wrapping within each document rather than the document set as a whole.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: I think the two parts should be redundant, as even where we have a hidden fact, the element for it in the hidden section gets the `xbrl-selected` class added, but as the contained code will behave unpredictably if either condition is not met it seems reasonable to explicitly check both.  I'll add a comment to the effect that the check is hopefully redundant.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: Yes, I'll rename it.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: I've now found a problem with the current approach, which is that we filter out everything with the `ixbrl-continuation` class, but it's possible for a table element to be both a continuation and a fact, and this filters those out.  I'll implement your suggestion of a list of all IDs, as we can then avoid including continuation IDs at all.

sagesmith-wf (approved):

brettkail-wk (dismissed):

brettkail-wk (approved):

PR #283, opened 2022-02-09
Fix same-document links

sagesmith-wf:
### Description
- Fixes same document hyperlinking
    -  Master's behavior was that a new tab was opened and the link didn't take you to the location within the document
    
### Steps to Test
1. Create an ixbrl file with a link to a different location within itself (preferrably a different page so you can verify it actually scrolls)
2. Generate an ixbrl-viewer using master
    a. Verify that clicking on the hyperlink opens a new tab and doesn't take you to the desired location within the doc
3. Generate an ixbrl-viewer using this branch
    a. Verify that clicking on the hyperlink doesn't open a new tab and navigates you within the document
  
  
### please review:
@paulwarren-wk @derekgengenbacher-wf @austinmatherne-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -152,7 +152,7 @@ Viewer.prototype._updateLink = function(n) {
                 e.preventDefault(); 
             });
         }
-        else {
+        else if (file) {
             // Open target in a new browser tab.  Without this, links will
             // replace the contents of the current iframe in the viewer, which
             // leaves the viewer in a confusing state.


iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -152,7 +152,7 @@ Viewer.prototype._updateLink = function(n) {
                 e.preventDefault(); 
             });
         }
-        else {
+        else if (file) {
```
brettkail-wk, line 155: In theory, this function should do true URL resolution (e.g., `new URL(url, current)`) before looking up the filename.  In practice, same-document reference uses empty string (so, `#id`), so we just check for empty `file` here.  In any case, let the browser do the navigation since the document is already loaded.  I'm not sure if `showElement` should be used (@paulwarren-wk?), but this at least restores the previous behavior.

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: I don't think `showElement` is needed, as we can just let the browser do its thing.

paulwarren-wk (approved):

derekgengenbacher-wf:
QA +10 Unit and Skynet tests pass successfully. Builds successfully. Built locally and tested locally no regressions found.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #294, opened 2022-04-10
Ensure that <script> tags always use default namespace (#282)

### LINKED ISSUE #282: Error on load caused by unrecognized script tag
**Important:**

I'm using the 1.0.3 release because it seems that newer versions do not work well with ESEF reports (but that's another issue).  
I don't think the version difference is related with my current problem but it's worth mentioning.

---

#### Description

In certain conditions opening the viewer gives a `Could not find viewer data` error.  
Unfortunately I was not able to determine the root cause of the issue nor can I provide the relevant iXBRL instance, because it is an actual financial report which may not be public yet.

image

However I found out that the data is not detected because the <script> tag enclosing the JSON data is actually prefixed with the xhtml namespace:

image

I thought this was because this prefix is explicitly defined in the report...

image

... but I was not able to reproduce the problem that way.

(_Edit_: I tried creating a viewer after removing that xhtml namespace prefix declaration from the original report and it worked, so I guess that is the problem after all.  
Still this case could be addressed.)

I suppose the problem could be fixed either when the viewer is created in Arelle...

image

... or in the javascript during the lookup for the `<script type="application/x.ixbrl-viewer+json">` (maybe by looking at the `localName` rather than the `tagName`?):

image

image

---

#### Miscellaneous

I mentionned two people in the issue for review (because the last one I wrote is still open ;) ), I hope that's ok.

**review**:
@derekgengenbacher-wf @sagesmith-wf

paulwarren-wk:
Certain input documents could cause the generated viewer to use prefixed `<script>` tags (`<xhtml:script>` rather than `<script>`).  Doing so will either cause the viewer to load completely, or the viewer will load, then fail with "Could not find viewer data") (the failure mode depends on whether the browser treats the file as HTML or XHTML).

The arrangement that triggers the bug is where the input document uses a default HTML5 namespace, but also declares a non-default prefix for that namespace, and the non-default NS declaration appears first, i.e.:

```
<html 
  xmlns:xhtml="http://www.w3.org/1999/xhtml"  
  xmlns="http://www.w3.org/1999/xhtml"
  >
```

A real world example of this is [this filing](https://filings.xbrl.org/549300CSLHPO6Y1AZN37/2021-12-31/ESEF/SE/0/)

This PR fixes this, so that the inserted `<script>` tags always use the default namespace.

Fixes #282

### DIFF
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -300,17 +300,21 @@ class IXBRLViewerBuilder:
             if child.tag == '{http://www.w3.org/1999/xhtml}body':
                 child.append(etree.Comment("BEGIN IXBRL VIEWER EXTENSIONS"))
 
-                e = etree.fromstring("<script xmlns='http://www.w3.org/1999/xhtml' type='text/javascript' />")
+                # Insert <script> tags, and make sure that they are in the
+                # default namespace, so that browsers in HTML mode will find
+                # them.
+                nsmap = { None: "http://www.w3.org/1999/xhtml" }
+                e = etree.SubElement(child, "{http://www.w3.org/1999/xhtml}script", nsmap = nsmap)
+                e.set("type", "text/javascript")
                 e.set("src", scriptUrl)
                 # Don't self close
                 e.text = ''
-                child.append(e)
 
                 # Putting this in the header can interfere with character set
-                # auto detection
-                e = etree.fromstring("<script xmlns='http://www.w3.org/1999/xhtml' type='application/x.ixbrl-viewer+json'></script>")
+                # auto detection due to its length
+                e = etree.SubElement(child, "{http://www.w3.org/1999/xhtml}script", nsmap = nsmap)
+                e.set("type", "application/x.ixbrl-viewer+json")
                 e.text = taxonomyDataJSON
-                child.append(e)
                 child.append(etree.Comment("END IXBRL VIEWER EXTENSIONS"))
                 break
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Security Insights
<details>
</details>
<details>
<li>Watched keyword <code>etree</code> in <code>iXBRLViewerPlugin/iXBRLViewer.py</code> line(s) ['307', '315'] added</li>
<li>Watched keyword <code>etree</code> in <code>iXBRLViewerPlugin/iXBRLViewer.py</code> line(s) ['303', '311'] removed</li>
</details>

## Action Items
---

brettkail-wk (approved):

austinmatherne-wk (approved):

sagesmith-wf (approved):

derekgengenbacher-wf (approved):
security +1

oliverjensen-wk (approved):
Security +1

There are no security concerns around namespaces in markup. Best I can tell, their purpose is entirely to inform the browser how to interpret `<something>` elements; previously, browsers would fail to recognize something that should be a script as a script, and this PR ensures that they do. Namespaces do not affect the internals of javascript execution, the SOP, or anything like that.

derekgengenbacher-wf:
QA +10 Unit and Skynet tests pass successfully. Builds successfully. Ran viewer on suggested doc which produced an error. Ran the suggested doc against this PR and the doc successfully ran the ixbrl-viewer.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #316, opened 2022-08-14
Fix scroll-to-fact with scaled content

paulwarren-wk:
This PR fixes the logic used to scroll the browser window (if necessary) whenever a fact is focused (e.g. using the next/prev tag buttons, or when clicking in search results).

[This ESEF filing](https://filings.xbrl.org/R0MUWSFPU8MPRO8K5P83/2021-12-31/ESEF/FR/0/) lays out each page at 10 times its intended size, and then scales it down to the correct size using a CSS transform.  This broke the previous scroll-to-fact logic, resulting in the browser often scrolling to a different page when attempting to focus on a fact.

This PR restores the intended functionality, which is that if the target fact is already fully visible, it will be focused without scrolling.  Otherwise, the browser will scroll to put the target fact in the center of viewport.

A simpler test document is available in `samples/src/scrollable-div`.

Unfortunately, highlighting of the selected fact in the above document still does not work correctly, as the highlight is 1px outline, which gets transformed to an invisible 0.1px border.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -378,11 +378,57 @@ Viewer.prototype.selectPrevTag = function (currentFact) {
     this._selectAdjacentTag(-1, currentFact);
 }
 
+// Calculate the intersection of two rectangles
+Viewer.prototype.intersect = function(r1, r2) {
+    const r3 = {
+        left: Math.max(r1.left, r2.left),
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Security Insights
<details>
</details>

## Action Items
---

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

brettkail-wk (approved):

PR #321, opened 2022-08-29
XT-1977: Fix nested .ixbrl-element hover outlines

aaroncameron-wk:
Exploit properties of inline-block and flex display to keep consistent line height regardless of element nesting

### Steps to Test:

Generate iXBRL viewer from document (has nested tags, ideally with a nested tag that spans more than one line, has paragraph text above and below to ensure consistent line heights). Can provide example doc on request. Ensure consistent hover/selection box sizing in individual and nested tagged elements.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/less/viewer.less
+++ b/iXBRLViewerPlugin/viewer/src/less/viewer.less
@@ -47,6 +47,16 @@ span {
   }
 }
 
+span {
+  &.ixbrl-element {
+    display: inline-block;
+
+    & > * {
+      display: flex;
+    }
+  }
+}
+
 .ixbrl-element {
   cursor: pointer;
 }


aviary-wf:
## Security Insights
<details>
</details>

## Action Items
---

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/less/viewer.less
```diff
@@ -41,9 +41,21 @@ span {
...
-    outline: dashed 2px @primary-focus;
-    outline-offset: 1px;
+  &.ixbrl-element {
+    display: inline-block;
+
+    & > [contextref] {
+      display: flex;
+    }
+
+    & > ix\:continuation {
```
brettkail-wk, line 51: I'm nervous that this requires `xmlns:ix`, but someone might choose to generate without (e.g., edit an existing iXBRL doc to use `xmlns:ixbrl` and `ixbrl:` throughout).  I think the JS adds a class, so just check `.ixbrl-continuation` if that's required?

iXBRLViewerPlugin/viewer/src/less/viewer.less
```diff
@@ -41,9 +41,21 @@ span {
...
   }
 
-  &.ixbrl-element:hover:not(.ixbrl-selected) {
-    outline: dashed 2px @primary-focus;
-    outline-offset: 1px;
+  &.ixbrl-element {
+    display: inline-block;
+
+    & > [contextref] {
+      display: flex;
```
brettkail-wk, line 48: I'm not sure what this is intended to do, but I worry that `[contextref]` is not selecting footnotes but should.

I don't understand CSS well enough to know the effects of `display:flex`, but I'm nervous it will have unintended side effects on document content.

aaroncameron-wk: The issue is caused when we have a `span.ixbrl-element > ix:* > span.ixbrl-element` structure. The `ix.*` element is considered to have 0px of height, which causes the flattening of the outline. Setting the `inline-block` value on `.ixbrl-element` fixes this to some extent:
Screen Shot 2022-08-29 at 1 24 16 PM

But the `ix.*` element is still causing some whitespace below the text which distorts the outline. The first thing I tried to fix that was setting line-height. That corrected the whitespace within the outline but also adjusted the entire line's height:
Screen Shot 2022-08-29 at 1 24 59 PM

`display: flex` was the only way I could find that got rid of that whitespace without effecting the overall line height:
Screen Shot 2022-08-29 at 2 47 37 PM
If the `span.ixbrl-element > ix:*` pattern appears in places that aren't paragraph/inline text then it could definitely cause problems. 

Maybe the first screenshot is satisfactory? It does have the benefit of having a clearer separation between the two outlines.

iXBRLViewerPlugin/viewer/src/less/viewer.less
aaroncameron-wk: Good point, `.ixbrl-continuation` can be used to the same affect.

iXBRLViewerPlugin/viewer/src/less/viewer.less
paulwarren-wk: Please could you provide an example document?  I can't immediately reproduce this issue with my own test doc.

[This method](https://github.com/Workiva/ixbrl-viewer/blob/master/iXBRLViewerPlugin/viewer/src/js/viewer.js#L115) is relevant here:

```javascript
    var wrapper = "<span>";
    var nn = n.getElementsByTagName("*");
    for (var i = 0; i < nn.length; i++) {
        if($(nn[i]).css("display") === "block") {
            wrapper = '<div>';
            break;
        }
    }
```

If the `.xbrl-element` wrapper is a `<span>` then it's because none of the descendants are `display: block`, but as it stands, the CSS is also applying this to `div.ixbrl-element` which means that it won't just be paragraph / inline text.

```css
& > [contextref] {
```
I think this can be simplified to `& > *` to avoid the footnote issue, although as above, if we do this, I think we want to constrain it to just the `span` wrappers.

I agree with @brettkail-wk that we need to be careful with this type of change given some of the weird and wonderful CSS that we see in real world ESEF documents.

aaroncameron-wk: [ixbrl_viewer.zip](https://github.com/Workiva/ixbrl-viewer/files/9454155/ixbrl_viewer.zip)
👍 on all of the above. It simplifies the changes quite a bit.

paulwarren-wk: Thanks, I now see what's causing this.  The issue here is that `div.pageView` sets `font-size: 0`, and the desired 12pt font-size is only actually set on the `<span>` element for each block of text.  For normal, non-nested tags, the `ix:nonNumeric` is a child of the `<span>` so the correct font size is in effect, but where we have nested tags, the `ix:nonNumeric` wraps the relevant `<span>` elements, and the font size in effect is `0`, resulting in the broken outline:

image

I assume that there's some good reason for this approach to font-size, although removing the `font-size: 0` on the `div` appears to have no effect on the document's appearance apart from fixing the outline.

I think the proposed change on this PR can be further simplified down to:

```css
span.ixbrl-element > * {
    display: inline-flex;
}
```

But, it's not without drawbacks:

1. The outline ends up with less vertical padding: 

Normal outline:

image

With `font-size: 0`, `display: inline-flex`

image

Obviously this is better than the broken outline, but this change will affect the appearance of all inline tags.

The appearance can be improved a bit by increasing `outline-offset`, but this also increases the horizontal padding:

image

2. For reasons that I don't understand, `display: inline-flex` also seems to reduce the width of the outline on [some documents](https://filings.xbrl.org/213800LRO7NS5CYQMN21/2022-03-31/ESEF/GB/1/213800LRO7NS5CYQMN21-2022-03-31/reports/ixbrlviewer.html#f-fact-21):

Before: 
image

After: 
image

The fact that the outline is thin in the first place is because that document is using a scaling transform and then a correspondingly inflated font size (see #316).  I've been separately working on a fix for the thin outline issue, and as part of that I've been wondering if we should take a fundamentally different approach and draw the outlines using JavaScript and `getClientRects()`.  I suspect that this would also address this issue, but it's a big change.

It's also worth nothing that the `font-size: 0` thing also breaks the "highlight XBRL elements" feature, and the use of `display: inline-flex` doesn't fix that:

image

I remain a bit nervous about what unwanted side-effects `display: flex` or `display: inline-flex` might have, but I've not been able to come up with any specific problematic examples.

aaroncameron-wk: The simplification you suggested does have a slightly different effect:
PR as-is:
Screen Shot 2022-08-31 at 8 11 48 AM

With suggested `inline-flex` simplification:
Screen Shot 2022-08-31 at 8 10 08 AM

The `getClientRects()` approach does seem like it would end up simplifying the outlining by decoupling it from the DOM structure and display properties. With that in mind, this PR would be more of a temporary solution to hold over until then, but is it worth the risk of causing issues with currently unforeseen use cases?

paulwarren-wk: > The simplification you suggested does have a slightly different effect:

Strange.  I'm sure there's a good reason, but I don't know what it is!

>  is it worth the risk of causing issues with currently unforeseen use cases?

Sadly I think I have now found the problem with the approach on this PR.  If you have a tag that spans more than one line, and is actually a single inline element then converting it to `inline-block` (or `inline-flex`) alters the layout.

Before:

image

After:

image

aaroncameron-wk:
This bug will instead be addressed by [TRANS-13139](https://jira.atl.workiva.net/browse/TRANS-13139)

PR #325, opened 2022-09-27
Fix highlighting of absolutely positioned elements within text tags

### LINKED ISSUE #139: Highlighting of absolutely positioned children of ix:nonNumeric elements
Hi,

tagged textblock can not be highlighted in iXBLR previewer.

I have created two examples:
[V1.zip](https://github.com/Workiva/ixbrl-viewer/files/5232141/V1.zip) uses css classes. Highlighting doesn't work.
If I remove the property "class='...' ", it works

[V2.zip](https://github.com/Workiva/ixbrl-viewer/files/5232153/V2.zip) has the same document but don't uss css classes.

If I remove the property **position:absolute;** in css, it works as expected.

Can someone help ?

Kind regards,
Nachtinho

paulwarren-wk:
If an `<ix:nonNumeric>` element contains absolutely positioned elements, those elements will not be highlighted by either fact selection (blue outline) or background color if "highlight all facts" is used.  Further, if the only content of the `<ix:nonNumeric>` element is absolutely positioned elements, this will lead to an outline around an empty element:

image

(note small square in top left of each paragraph)

This PR fixes both of these issues, by applying separate highlighting to absolutely positioned sub elements, and removing highlighting from any element with zero height.

The issue addressed is likely to become common with the introduction of block tagging on ESEF documents that have been converted from PDF, and individual lines are often absolutely positioned `<div>` elements.

The end result is a bit messier than on a "normal" document, where a tagged paragraph would have a single border, but it would be hard to do better:

image

The issue can be seen with the sample on #138, "testdoc" in samples, and in
[this report](https://github.com/Workiva/ixbrl-viewer/files/9659519/7437009IVYWGEE4S7B77-2021-12-31-en.zip)

Fixes #139

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -196,6 +196,8 @@ Viewer.prototype._findOrCreateWrapperNode = function(domNode) {
     /* Otherwise, insert a <span> as wrapper */
     if (node.length == 0) {
         node = this._wrapNode(domNode);
+        // Create a node set of current node and all absolutely positioned
+        // descendants.
         node = node.find("*").addBack().filter(function () {
             return (this == node[0] || $(this).css("position") == "absolute");
         });
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -648,14 +648,14 @@ Viewer.prototype.highlightAllTags = function (on, namespaceGroups) {
     var viewer = this;
     if (on) {
         $(".ixbrl-element", this._contents).each(function () {
-            const factId = $(this).data('ivid')[0];
-            const ixn = viewer._ixNodeMap[factId];
-            if (!ixn.isContinuation) {
-                const elements = viewer.elementsForItemIds([factId].concat(ixn.continuationIds()));
+            // Find the first ixn for this element that isn't a continuation.
+            const ixn = $(this).data('ivid').map(id => viewer._ixNodeMap[id]).filter(ixn => !ixn.isContinuation)[0];
+            if (ixn != undefined) {
+                const elements = viewer.elementsForItemIds([ixn.id].concat(ixn.continuationIds()));
                 elements.addClass("ixbrl-highlight");
 
-                if (!ixn.footnote && !ixn.isContinuation) {
-                    const i = groups[report.getItemById(factId).conceptQName().prefix];
+                if (!ixn.footnote) {
+                    const i = groups[report.getItemById(ixn.id).conceptQName().prefix];
                     if (i !== undefined) {
                         elements.addClass("ixbrl-highlight-" + i);
                     }
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -181,28 +181,28 @@ Viewer.prototype._findOrCreateWrapperNode = function(domNode) {
     const v = this;
     /* Is the element the only significant content within a <td> or <th> ? If
      * so, use that as the wrapper element. */
-    var node = $(domNode).closest("td,th").eq(0);
+    var nodes = $(domNode).closest("td,th").eq(0);
     const innerText = $(domNode).text();
-    if (node.length == 1 && innerText.length > 0) {
+    if (nodes.length == 1 && innerText.length > 0) {
         // Use indexOf rather than a single regex because innerText may
         // be too long for the regex engine 
-        const outerText = $(node).text();
+        const outerText = $(nodes).text();
         const start = outerText.indexOf(innerText);
         const wrapper = outerText.substring(0, start) + outerText.substring(start + innerText.length);
         if (/[0-9A-Za-z]/.test(wrapper)) {
-            node = $();
+            nodes = $();
         } 
     }
     /* Otherwise, insert a <span> as wrapper */
-    if (node.length == 0) {
-        node = this._wrapNode(domNode);
+    if (nodes.length == 0) {
+        nodes = this._wrapNode(domNode);
         // Create a node set of current node and all absolutely positioned
         // descendants.
-        node = node.find("*").addBack().filter(function () {
-            return (this == node[0] || $(this).css("position") == "absolute");
+        nodes = nodes.find("*").addBack().filter(function () {
+            return (this == nodes[0] || $(this).css("position") == "absolute");
         });
     }
-    node.each(function (i) {
+    nodes.each(function (i) {
         if (this.getBoundingClientRect().height == 0) {
             $(this).addClass("ixbrl-no-highlight"); 
         }
@@ -213,7 +213,7 @@ Viewer.prototype._findOrCreateWrapperNode = function(domNode) {
             $(this).addClass("ixbrl-sub-element"); 
         }
     });
-    return node;
+    return nodes;
 }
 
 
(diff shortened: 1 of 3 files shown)


aviary3-wk:
## Security Insights
<details>
</details>

## Action Items
---

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -196,15 +197,59 @@ Viewer.prototype._findOrCreateWrapperNode = function(domNode) {
     /* Otherwise, insert a <span> as wrapper */
     if (node.length == 0) {
         node = this._wrapNode(domNode);
+        node = node.find("*").addBack().filter(function () {
```
brettkail-wk, line 200: I'm admittedly quite new to jQuery, but a small comment would have helped:
```
/* Select the node and all absolutely-positioned descendants. */
```
Also, I wonder if we should rename this to `nodes` as you did in `_preProcessiXBRL`?

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -486,7 +531,13 @@ Viewer.prototype.selectElementByClick = function (e) {
     var itemIDList = [];
     var viewer = this;
     var sameContentAncestorId;
-    e.parents(".ixbrl-element").addBack().each(function () { 
+    // If the user clicked on a sub-element, treat as if we clicked the first
+    // non-sub-element ancestor in the DOM hierarchy - which would typically be
+    // the corresponding ixbrl-element.
+    if (e.hasClass('ixbrl-sub-element')) {
+        e = e.parents('.ixbrl-element:not(.ixbrl-sub-element)').first();
```
brettkail-wk, line 538: Out of curiosity, what happens for nested facts?  It seems like it would be possible for the nested fact region to have both ixbrl-sub-element and ixbrl-element, and then it wouldn't be possible to click on the nested element.  If it's not a problem, can you expand the comment to explain why this logic doesn't cause that problem?

paulwarren-wk:
Reverted to draft as I've found a bug that needs investigating.

```
Uncaught TypeError: Cannot read properties of undefined (reading '0')
    at ./src/js/viewer.js.Viewer._selectAdjacentTag (viewer.js:387:1)
    at ./src/js/viewer.js.Viewer.selectNextTag (viewer.js:423:1)
    at HTMLSpanElement.<anonymous> (inspector.js:133:6)
    at HTMLSpanElement.dispatch (jquery.js:5429:1)
    at elemData.handle (jquery.js:5233:1)
```
On https://www.sec.gov/ix?doc=/Archives/edgar/data/1772921/000114036122028384/brhc10040150_10q.htm

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: Additional comments added.

The subtlety here is that `.ixbrl-sub-element` gets applied directly to elements in the source document, whereas `.ixbrl-element` gets added to wrapper nodes that are inserted by the viewer, so we can ignore `.ixbrl-sub-element` elements without missing a real tag.  

Note that this particular line is only needed for the "same content ancestor" detection in the following traversal.

There is one possible edge case: for tags in table cells we sometimes use the enclosing `<td>` or `<th>` as the `ixbrl-element` rather than using a wrapper.  If the table cell were to be absolutely positioned, and to be within another text tag, then we'd have an element that would be both an `ixbrl-element` and an `ixbrl-sub-element`.  In this case I think we would miss the nested table cell tag.  This seems unlikely, although PDF-converted HTML never ceases to surprise :) 

I think we could address it by inverting the tests and checking for the "real" element classes (`.ixbrl-element-nonfraction` etc.) rather than the absence of `.ixbrl-sub-element`.

brettkail-wk: > If the table cell were to be absolutely positioned, and to be within another text tag, then we'd have an element that would be both an ixbrl-element and an ixbrl-sub-element. In this case I think we would miss the nested table cell tag. This seems unlikely

This scenario seems plausible for a text block surrounding a table.  I believe Workiva currently uses `<td><div absolutely-positioned-style>`, but we have considered switching to `<td absolutely-positioned-style>`.  I think we should try to not break that use-case if possible.

brettkail-wk: > I think we could address it by inverting the tests and checking for the "real" element classes (.ixbrl-element-nonfraction etc.) rather than the absence of .ixbrl-sub-element.

Would that cause problems for continuations in tables nested within text blocks?  Perhaps that's unlikely enough that we can use that approach and just add a comment that it might cause problems.

paulwarren-wk: > Would that cause problems for continuations in tables nested within text blocks?

I think that should be OK, as we can include `.ixbrl-continuation` in the list of real element classes.

The problem with the approach is that we can end up with the same fact twice in `itemIDList` and thus appearing twice in the inspector, as we'll get the same ID from the real `.ixbrl-element` and the `.ixbrl-sub-element`, so we need to deduplicate `itemIDList`.

I'll create some test cases for this, and then attempt to fix it.

paulwarren-wk:
> I've found a bug that needs investigating.

Now fixed in a8a3ce1

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: Done in 544b675 and 9ee05af.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: This is done in 74a99a557cacb1380ae3d04893a61ad5e55ff212.

I've realised that there's actually no need to store IDs in `ivid` on sub-elements, as we always get the list of IDs from the ancestor `.ixbrl-element`s anyway.  This removes the need to de-duplicate.

I've also reworked the classes, so that sub elements are no longer also given the `ixbrl-element` class (unless they are also `ixbrl-element` by virtue of being a `<td>` or `<th>` acting as a wrapper).

brettkail-wk (dismissed):

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -572,22 +647,24 @@ Viewer.prototype.highlightAllTags = function (on, namespaceGroups) {
...
-            elements.addClass("ixbrl-highlight");
-
-            if (!ixn.footnote) {
-                var i = groups[report.getItemById(factId).conceptQName().prefix];
-                if (i !== undefined) {
-                    elements.addClass("ixbrl-highlight-" + i);
+        $(".ixbrl-element", this._contents).each(function () {
+            const factId = $(this).data('ivid')[0];
+            const ixn = viewer._ixNodeMap[factId];
+            if (!ixn.isContinuation) {
```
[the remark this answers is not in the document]
paulwarren-wk: Ah - it wasn't intended to be the same logic, but it was broken.  The same element could be both a "primary" iXBRL element, and a continuation, so we shouldn't filter out all elements that have the `ixbrl-continuation` class - although it would still have been highlighted as the primary element for the continuation also highlights its continuations (unless that primary element was also sharing a table cell with a continuation...)

The updated code uses the first non-continuation element to determine the color of the highlighting which is still an arbitrary choice, but I think preferable to basing it on a continuation.

I think we need to revisit the highlighting of nested tags anyway (see #326).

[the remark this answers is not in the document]
brettkail-wk: The new logic makes much more sense to me, thanks.

brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
QA +10 Builds successfully. Testing locally succeeded.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #330, opened 2022-10-05
Handle continuations for items in non-default target document

paulwarren-wk:
This PR address a bug encountered where a document contains `ix:continuation` elements for facts or footnotes that are not in the default target document.   The viewer should ignore facts that are not in the default target document, but continuations for such elements were not being ignored (they don't have a `target` attribute so they can't be trivially ignored).  This lead to an exception when they were clicked on, or when they were selected using next/prev tag.

The bug can be seen in [this filing](https://filings.xbrl.org/5299004SWFL5JAN4W830/2022-06-30/ESEF/DK/0/5299004SWFL5JAN4W830-2022-06-30-en/reports/ixbrlviewer.html)

Open the viewer, press the "next tag" button and note that "June 30, 2022" is highlighted, but there are no fact details in the inspector, and we have an exception on the console.  Similarly, clicking on this text causes an exception.

This PR finds all orphaned continuation elements, and removes them so that they can't be clicked on, and are not included in the next/prev tag list.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -27,7 +27,7 @@ export function Viewer(iv, iframes, report) {
     this.onMouseLeave = $.Callbacks();
 
     this._ixNodeMap = {};
-    this._continuedAtMap = {};
+    this._continuationMap = {};
     this._docOrderIDIndex = [];
 }
 
@@ -69,17 +69,17 @@ function localName(e) {
 }
 
 Viewer.prototype._buildContinuationMap = function() {
-    var continuations = Object.keys(this._continuedAtMap)
+    var continuations = Object.keys(this._continuationMap)
     for (var i = 0; i < continuations.length; i++) {
         var id = continuations[i];
-        if (this._continuedAtMap[id].isFact) {
+        if (this._continuationMap[id].isHead) {
             var parts = [];
             var nextId = id;
-            while (this._continuedAtMap[nextId].continuedAt !== undefined) {
-                nextId = this._continuedAtMap[nextId].continuedAt;
+            while (this._continuationMap[nextId].continuedAt !== undefined) {
+                nextId = this._continuationMap[nextId].continuedAt;
                 if (this._ixNodeMap[nextId] !== undefined) {
-                    this._continuedAtMap[nextId] = this._continuedAtMap[nextId] || {};
-                    this._continuedAtMap[nextId].continuationOf = id;
+                    this._continuationMap[nextId] = this._continuationMap[nextId] || {};
+                    this._continuationMap[nextId].continuationOf = id;
                     parts.push(this._ixNodeMap[nextId]);
                 }
                 else {
@@ -88,6 +88,29 @@ Viewer.prototype._buildContinuationMap = function() {
                 }
             }
             this._ixNodeMap[id].continuations = parts;
+        }
+    }
+    // The document may contain ix:continuations for facts or footnotes that
+    // are not in the default target document, which we need to ignore.
+    // We can't ignore these up-front, because ix:continuation doesn't have a
+    // target attribute - we need to first resolve the continuation chain.
+    // This strips all continuation elements in the map that we haven't found
+    // the head item for.
+    for (const [id, item] of Object.entries(this._continuationMap)) {
+        if (!item.isHead && item.continuationOf === undefined) {
+            const node = this._ixNodeMap[id].wrapperNodes.first();
+            const ivids = node.data('ivid').filter(item => item != id);
+            node.data('ivid', ivids);
+            if (ivids.length == 0) {
+                node.removeClass("ixbrl-continuation").removeClass("ixbrl-element");
+                // Remove the ixbrl-sub-element class from any sub elements
+                // that no longer have a ixbrl-element ancestor.
+                node.find(".ixbrl-sub-element").each(function () {
+                    if ($(this).parents().filter('.ixbrl-element').length == 0) {
+                        $(this).removeClass("ixbrl-sub-element").removeClass("ixbrl-continuation");
+                    }
+                });
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

brettkail-wk (dismissed):

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -90,6 +90,29 @@ Viewer.prototype._buildContinuationMap = function() {
...
+    // target attribute - we need to first resolve the continuation chain.
+    // This strips all continuation elements in the map that we haven't found
+    // the head item for.
+    for (const [id, item] of Object.entries(this._continuationMap)) {
+        if (!item.isHead && item.continuationOf === undefined) {
+            const node = this._ixNodeMap[id].wrapperNodes.first();
+            const ivids = node.data('ivid').filter(item => item != id);
+            node.data('ivid', ivids);
+            if (ivids.length == 0) {
+                node.removeClass("ixbrl-continuation").removeClass("ixbrl-element");
```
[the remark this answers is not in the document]
paulwarren-wk: Yes, I was having exactly the same thought as I wrote it, and the clean-up logic became more and more involved.

I was thinking that the first pass would be purely to build a continuation map, so we can identify which target document(s) a continuation corresponds to in the second pass, but it might make sense to do more, as you suggest.  I'll look into this.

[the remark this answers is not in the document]
paulwarren-wk: I've now done this, and the result is much cleaner.  continuations now store the ID of their head node(s) which cleans up other parts of the code.

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -244,102 +254,96 @@ Viewer.prototype._addIdToNode = function(node, id) {
...
-// attribute(s) of corresponding IX element(s).  i.e. facts have fact IDs,
-// continuations have continuation IDs, footnotes have footnote IDs.  It's
-// possible for it to be a mix of different types.
+// attribute(s) of corresponding IX item(s).  Continuations have the IDs of
+// their head items (fact or footnotes).
+// "ivid" can be a mix of different types.
 //
 // Viewer._ixNodeMap is a map of these IDs to IXNode objects
 //
 // Viewer._docOrderIDIndex is a list of fact and footnote IDs in document
```
brettkail-wk, line 263: I think it's now a list of {id,document} tuples, right?  (I don't JavaScript much, but I wonder if it would be clearer with an explicit DocumentScopedId class or something.)

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -244,102 +254,96 @@ Viewer.prototype._addIdToNode = function(node, id) {
...
-        if (id !== null) {
-            nodes = $(n);
-            nodes.addClass("ixbrl-element").data('ivid', [id]);
-            this._docOrderIDIndex.push(id);
-            /* We may have already seen the corresponding ix element in the hidden
-             * section */
+
+            // For a continuation, store the IX ID of the item, not the continuation
+            const headId = isContinuation ? this.continuationOfMap[id] : id;
+            this._addIdToNode(nodes.first(), headId);
```
brettkail-wk, line 287: It looks like `continuationOfMap` is an map<continuationId,list<primaryId>>, so `headId` is now either a single id or an array, but it looks like `_addIdToNode` wants a single id.  Is this working as expected?

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: I've refactored the index itself into a separate class, which I think makes it more readable overall.  I've previously been reluctant to break out new classes because the pre-ES6 class syntax was so clunky, but it's probably about time we switched to using `class`.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: Good spot.  Amazingly, this does appear to be working as intended (by fluke), because apparently in JS:

```javascript
> arr = ["a","b","c"];
(3) ['a', 'b', 'c']
> arr[1];
'b'
> arr[[1]];
'b'
```

I've no idea why this array flattening happens, but have fixed this code in 1adcede.

brettkail-wk: I know keys are `.toString()`'ed before being used as keys, but presumably if there ever _were_ multiple keys that things wouldn't work?  I guess the scenario is two fact targeting the same continuation.  If we don't already have a test for that (e.g., manual), it might be valuable to add one.

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: Yes, I should have said "it's not as obviously broken as you might expect" :-)  

This is intended to deal with a continuation that is shared between two facts, but it turns out that that's actually [illegal](https://www.xbrl.org/specification/inlinexbrl-part1/rec-2013-11-18/inlinexbrl-part1-rec-2013-11-18.html#d1e1266):

> If the continuedAt attribute is present, the {continuation reference} property MUST NOT match the {continuation reference} property of any other element in the [Inline XBRL Document Set](https://www.xbrl.org/specification/inlinexbrl-part1/rec-2013-11-18/inlinexbrl-part1-rec-2013-11-18.html#term-inlinexbrldocumentset).

I'll remove this, and replace it with a check that we don't have shared continuations.

paulwarren-wk: Done in ff4f118

brettkail-wk (dismissed):

austinmatherne-wk (approved):

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -715,11 +704,8 @@ Viewer.prototype.showDocumentForItemId = function(itemId) {
     this.selectDocument(this._ixNodeMap[itemId].docIndex);
 }
 
-Viewer.prototype.currentDocument = function () {
```
brettkail-wk, line 718: @derekgengenbacher-wf is adding a usage of this method in #334.  Can you undo this removal (reimplement in terms of `_currentDocumentIndex`)?

paulwarren-wk: Done in 6d693f2.

austinmatherne-wk (approved):

austinmatherne-wk:
# Verified

* ix:continuation elements for facts not in the target document are ignored

QA +10
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #365, opened 2023-01-25
Show warning dialog on more than 10,000 continuation elements

### LINKED ISSUE #358: Viewer hanging with specifc ESEF iXBRL instance
### What happened?
When opening a specific ESEF iXBRL instance in the iXBRL Viewer, the iXBRL Viewer keeps loading forever when Chrome is used:

Chrome 108 hanging

In Firefox, the document is loaded, but at the end (at least for versions 1.1.31 and 1.1.38), the fact details are not correctly shown.

### What did we expect to happen?
The iXBRL Viewer displays correctly the document and the fact details are correctly shown.

In Firefox we observe:

Firefox 108

### ESEF instance exhibiting the issue

See [stef-2022-12-31-FR.zip](https://github.com/Workiva/ixbrl-viewer/files/10393030/stef-2022-12-31-FR.zip)

A generated iXBRL viewer file is [stef-2022-12-31-FR_ixbrlview.zip](https://github.com/Workiva/ixbrl-viewer/files/10393041/stef-2022-12-31-FR_ixbrlview.zip)

### Browser instances used for the tests

- Firefox 108.0.2: iXBRL Viewer usable after 53s
- MS Edge 108.0.1462.76: no reply after 600s
- Google Chrome 108.0.5359.125: no reply after 600s

### iXBRL Viewer versions tested:

1.1.38: document is loaded in Firefox, but selecting an XBRL fact does not work properly on the given instance, inline frame not tested.
1.1.31: same as 1.1.38, but fact selection works fine in inline frame (see screenshot).

Many thanks for your help,

Greg

paulwarren-wk:
This PR introduces a check on the number of continuation elements in a document, and offers to gracefully stop trying to load if there are more than 10,000.  

We now display the following dialog:

image

This dialog comes up fairly quickly, as the initial scan for continuation elements is fairly cheap.  

If you choose "Cancel" the loading mask is removed, but the inspector is grayed out:

image

This fixes #358, where a document with 70,000 continuation elements would cause the viewer to hang when loading, and would typically leave the browser tab unresponsive for several minutes.

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -51,7 +51,7 @@ Viewer.prototype._checkContinuationCount = function() {
             }
         });
     }
-    return new Promise(r => r());
+    return Promise.resolve();
 }
 
 Viewer.prototype.initialize = function() {
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -20,8 +20,6 @@ import { setDefault } from './util.js';
 import { DocOrderIndex } from './docOrderIndex.js';
 import { MessageBox } from './messagebox.js';
 
-const CONTINUATION_LIMIT = 10000;
-
 export class DocumentTooLargeError extends Error {}
 
 export function Viewer(iv, iframes, report) {
@@ -39,8 +37,8 @@ export function Viewer(iv, iframes, report) {
 }
 
 Viewer.prototype._checkContinuationCount = function() {
-        const continuationCount = Object.keys(this.continuationOfMap).length
-    if (continuationCount > CONTINUATION_LIMIT) {
+    const continuationCount = Object.keys(this.continuationOfMap).length
+    if (continuationCount > this._iv.options.continuationElementLimit) {
         const contents = $('<div></div>')
             .append($('<p></p>').text(`This document contains a very large number of iXBRL elements (found ${continuationCount} ix:continuation elements).`))
             .append($('<p></p>').text('You may experience performance problems viewing this document, or the viewer may not load at all.'))
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -18,6 +18,11 @@ import { escapeRegex } from './util.js'
 import { IXNode } from './ixnode.js';
 import { setDefault } from './util.js';
 import { DocOrderIndex } from './docOrderIndex.js';
+import { MessageBox } from './messagebox.js';
+
+const CONTINUATION_LIMIT = 10000;
+
+export class DocumentTooLargeError extends Error {}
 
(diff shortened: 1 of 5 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -18,6 +18,11 @@ import { escapeRegex } from './util.js'
 import { IXNode } from './ixnode.js';
 import { setDefault } from './util.js';
 import { DocOrderIndex } from './docOrderIndex.js';
+import { MessageBox } from './messagebox.js';
```
sagesmith-wf, line 21: Arbitrary line:

This is a great solution, thank you! There is a similar issue with spans when the document takes up 3gb in chrome, idk if we could get something similar where we check the memory used and give an error about the doc taking up too much memory, thoughts? (Probably not on this issue, but this solution inspired me)

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -18,6 +18,11 @@ import { escapeRegex } from './util.js'
...
+
+const CONTINUATION_LIMIT = 10000;
```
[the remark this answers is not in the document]
paulwarren-wk: Yes, we could include something in the report data block, although I think I'd prefer to add it as a "don't show document size warning" flag, rather than making the limit configurable.  I think it's likely that we'll want to add other size checks, and making them all independently configurable at a document-level seems unnecessary.

In any event, I should probably move this limit to the iXBRLViewer options object, so that it can be overridden in extensions.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: I'm sure we can do something.  I can't see any way to get the DOM memory usage from a browser, so we'd probably have to count the number of elements and use that as a proxy.

I was also planning to extend this to include a limit on the total number of facts.  From what I could tell, the thing that really kills the performance on this document is inserting the wrapper nodes, so we'll hit a similar problem with a large number of facts.

iXBRLViewerPlugin/viewer/src/js/viewer.js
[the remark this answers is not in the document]
sagesmith-wf: Awesome, all of that seems good to me

iXBRLViewerPlugin/viewer/src/js/viewer.js
sagesmith-wf: Awesome seems good to me, thank you!

sagesmith-wf (approved):

aaroncameron-wk (dismissed):

brettkail-wk (approved):

sagesmith-wf (approved):

austinmatherne-wk (approved):

rmconsole-wf (approved):
+1 from RM

Issue #370, opened 2023-02-01
[BUG] arelle gui save ixbrl viewer instance button doesn't work


sagesmith-wf:
### Description
I get [this error](https://github.com/Workiva/ixbrl-viewer/blob/8c6b0c05d1ddc5e7b98cb98cc479392c8d4ca4e2/iXBRLViewerPlugin/iXBRLViewer.py#L205) when trying to save an ixbrl-viewer instance via the arelle gui.

### Repro Instructions
1. Clicked `Tools -> Save iXBRL Viewer Instance` 
2. Set options to
  a. Script URL: `https://cdn-prod.wdesk.com/ixbrl-viewer/1.1.42/ixbrlviewer.js`
  b. iXBRL file: `viewer.html`
3. Hit OK

Changing the ixbrl-viewer code to set the buffer to an empty list when its none also doesn't seem to work, though it does make the error go away. Similarly using a dirty doc and going through the save messages file also doesn't seem to fix it

### Note:
On master arelle

### Version

1.1.42

### With which browsers are you experiencing the bug?

Chrome

### Documents

[workiva-8k.zip](https://github.com/Workiva/ixbrl-viewer/files/10560839/workiva-8k.zip)

### Screenshots

<img width="367" alt="Screenshot 2023-02-01 at 12 07 37 PM" src="https://user-images.githubusercontent.com/17255946/216141619-65ae017d-90ed-4cb6-bc65-9cc06be3be75.png">
<img width="742" alt="Screenshot 2023-02-01 at 12 07 52 PM" src="https://user-images.githubusercontent.com/17255946/216141637-021811c8-c06a-4730-89be-ddbcb675ca25.png">
<img width="517" alt="Screenshot 2023-02-01 at 12 08 01 PM" src="https://user-images.githubusercontent.com/17255946/216141651-e4eaf8f9-1be1-4a13-afbf-5b123e3c8aa3.png">

PR #371, opened 2023-02-01
Fix the save viewer functionality

sagesmith-wf:
#### Reason for change
#370

#### Description of change
- Starts logging on load (similar to edgar renderer [1](https://github.com/Arelle/EdgarRenderer/blob/60d4b56fcd2a79356724196664d2e56b5e8b1251/__init__.py#L12960), [2](https://github.com/Arelle/EdgarRenderer/blob/60d4b56fcd2a79356724196664d2e56b5e8b1251/__init__.py#L1565))
- Allows viewer to be saved after its loaded

#### Steps to Test
- Run on the filing from the issue

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/__init__.py
+++ b/iXBRLViewerPlugin/__init__.py
@@ -97,7 +97,7 @@ def iXBRLViewerMenuCommand(cntlr):
     dialog = SaveViewerDialog(cntlr)
     if dialog.accepted and dialog.filename():
         viewerBuilder = IXBRLViewerBuilder(modelXbrl)
-        iv = viewerBuilder.createViewer(scriptUrl=dialog.scriptUrl())
+        iv = viewerBuilder.createViewer(scriptUrl=dialog.scriptUrl(), showValidations=False)
         if iv is not None:
             iv.save(dialog.filename())
 
@@ -133,10 +133,12 @@ def viewMenuExtender(cntlr, viewMenu, *args, **kwargs):
     cntlr.launchIXBRLViewer.trace("w", setLaunchIXBRLViewer)
     erViewMenu.add_checkbutton(label=_("Launch viewer on load"), underline=0, variable=cntlr.launchIXBRLViewer, onvalue=True, offvalue=False)
 
+
 def guiRun(cntlr, modelXbrl, attach, *args, **kwargs):
     """ run iXBRL Viewer using GUI interactions for a single instance or testcases """
     if cntlr.hasGui and cntlr.launchIXBRLViewer.get() and modelXbrl.modelDocument.type in (Type.INLINEXBRL, Type.INLINEXBRLDOCUMENTSET):
         launchLocalViewer(cntlr, modelXbrl)
+
 
 __pluginInfo__ = {
     'name': 'ixbrl-viewer',
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -16,6 +16,7 @@ from arelle import XbrlConst
 from arelle.ModelDocument import Type
 from arelle.ModelValue import QName, INVALIDixVALUE
 from lxml import etree
+from copy import deepcopy
 import json
 import math
 import re
@@ -301,6 +302,7 @@ class IXBRLViewerBuilder:
                     if body_child.tag == '{http://www.w3.org/1999/xhtml}script' and body_child.get('type','') == 'application/x.ixbrl-viewer+json':
                         self.dts.error("viewer:error", "File already contains iXBRL viewer")
                         return False
+
                 child.append(etree.Comment("BEGIN IXBRL VIEWER EXTENSIONS"))
 
                 # Insert <script> tags, and make sure that they are in the
@@ -326,7 +328,6 @@ class IXBRLViewerBuilder:
         """
         Create an iXBRL file with XBRL data as a JSON blob, and script tags added
         """
-
         dts = self.dts
         iv = iXBRLViewer(dts)
         self.idGen = 0
@@ -351,17 +352,19 @@ class IXBRLViewerBuilder:
 
         if dts.modelDocument.type == Type.INLINEXBRLDOCUMENTSET:
             # Sort by object index to preserve order in which files were specified.
-            docSet = sorted(dts.modelDocument.referencesDocument.keys(), key=lambda x: x.objectIndex)
-            docSetFiles = list(map(lambda x: os.path.basename(x.filepath), docSet))
-            self.taxonomyData["docSetFiles"] = docSetFiles
+            xmlDocsByFilename = {
+                os.path.basename(doc.filepath): deepcopy(doc.xmlDocument)
+                for doc in sorted(dts.modelDocument.referencesDocument.keys(), key=lambda x: x.objectIndex)
+            }
+            self.taxonomyData["docSetFiles"] = list(xmlDocsByFilename.values())
 
-            for n in range(0, len(docSet)):
-                iv.addFile(iXBRLViewerFile(docSetFiles[n], docSet[n].xmlDocument))
+            for filename, xmlDocument in xmlDocsByFilename.items():
+                iv.addFile(iXBRLViewerFile(filename, xmlDocument))
 
-            xmlDocument = docSet[0].xmlDocument 
+            xmlDocument = next(iter(xmlDocsByFilename.values()))
 
         else:
-            xmlDocument = dts.modelDocument.xmlDocument
+            xmlDocument = deepcopy(dts.modelDocument.xmlDocument)
             filename = os.path.basename(dts.modelDocument.filepath)
             iv.addFile(iXBRLViewerFile(filename, xmlDocument))
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 .idea
 .vscode
-node_modules/*
+node_modules/
+iXBRLViewerPlugin/viewer/dist
+iXBRLViewerPlugin/
+*/__pycache__


aviary-wf:
## Security Insights

## Action Items
---

austinmatherne-wk (dismissed):

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -297,10 +297,16 @@ def addViewerToXMLDocument(self, xmlDocument, scriptUrl):
...
-                    if body_child.tag == '{http://www.w3.org/1999/xhtml}script' and body_child.get('type','') == 'application/x.ixbrl-viewer+json':
-                        self.dts.error("viewer:error", "File already contains iXBRL viewer")
-                        return False
+                    if body_child.tag == '{http://www.w3.org/1999/xhtml}script':
+                        if body_child.get('type','') == 'application/x.ixbrl-viewer+json':
+                            alreadyWritten = True
+                        elif body_child.get('type', '') == 'text/javascript' and 'ixbrlviewer.js' in body_child.get('src'):
+                            body_child.set('src', scriptUrl)
+                if alreadyWritten:
+                    return True
```
brettkail-wk, line 308: Out of curiosity, why this change?

iXBRLViewerPlugin/__init__.py
```diff
@@ -150,4 +158,5 @@ def guiRun(cntlr, modelXbrl, attach, *args, **kwargs):
     'CntlrWinMain.Menu.Tools': toolsMenuExtender,
     'CntlrWinMain.Menu.View': viewMenuExtender,
     'CntlrWinMain.Xbrl.Loaded': guiRun,
+    'ModelDocument.IsPullLoadable': guiStartLogging,
```
brettkail-wk, line 161: I'm not familiar with the Arelle plugin model.  It seems odd to do "real" work in an `Is` method; is this common practice, or is there a better even that can be used?  Does this need to stop logging at some point (i.e., does a long-running session keep accumulating messages in this handler)?  If you close/open repeatedly, does the plugin still work (i.e., new ModelXbrl and plugins are run against that)?

sagesmith-wf: Let me preface this with, I think you have raised some really good points, though there isn't a lot to go off of so I will just layout how I decided to do things and we can potentially come up with a better flow.

> It seems odd to do "real" work in an Is method; is this common practice, or is there a better even that can be used? 

I wouldn't say common, but of the one example I have to go off of someone using arelle similarly to the ixbrl-viewer (EdgarRenderer) I see this exact same thing being done. Herm added the same thing [1](https://github.com/Arelle/EdgarRenderer/blame/60d4b56fcd2a79356724196664d2e56b5e8b1251/__init__.py#L1565), [2](https://github.com/Arelle/EdgarRenderer/blame/60d4b56fcd2a79356724196664d2e56b5e8b1251/__init__.py#L1296-L1300) after making a change in arelle around the same time that [makes the buffer initialized to None](https://github.com/Arelle/Arelle/blame/fa0343705369ce6f0516ceafee993a53ee10eab0/arelle/CntlrWinMain.py#L1535-L1537).  Talking to Herm this morning, he confirmed that the EdgarRenderer was a good example to go off of, though also admitted the flow is a bit confusing so 🤷 . I will also point out that I think [failing because there is no log buffer is odd](https://github.com/Workiva/ixbrl-viewer/blob/7bcde691a5edd59487e9e3158554dffb14c7886c/iXBRLViewerPlugin/iXBRLViewer.py#L205) so perhaps just skipping and warning about the lack of validation messages not meaning your doc is necessarily clean would be better? The problem with that approach is that nothing in [arelle actually starts the buffer](https://github.com/Arelle/Arelle/blob/fa0343705369ce6f0516ceafee993a53ee10eab0/arelle/CntlrWinMain.py#L1535) so then there will never actually be validations and you will always get the warning

> Does this need to stop logging at some point (i.e., does a long-running session keep accumulating messages in this handler)? 

I haven't been able to make it crash yet but I suppose if you ran tons and tons of broken docs through it, it might crash so it'd probably be good to find an appropriate time to call stop, though I don't have a good example of when to call stop. The Edgar renderer seems to be doing it for [dubious reasons](https://github.com/Arelle/EdgarRenderer/blob/60d4b56fcd2a79356724196664d2e56b5e8b1251/__init__.py#L1444). I have the opposite fear that they have where I am worried about calling stop because everything shares that logger and calling stop will because the messages will randomly start going to something else in the middle of other operations potentially. Though maybe the same argument is true for calling `startLoggingBuffer`, though there was an error coded around this one.

> If you close/open repeatedly, does the plugin still work (i.e., new ModelXbrl and plugins are run against that)?

Yes

paulwarren-wk: Would `CntlrWinMain.Xbrl.Open` be suitable?

As far as I can see the log buffer does get appended to indefinitely.  I think it's probably necessary to clear the buffer when closing a document, as otherwise, validation messages relating to one document will be included against the next.  Also, if you run `Tools->Validation->Validate` then I suspect that this will cause validation messages to be doubled up.

I don't know how important it is to be able to included validation messages in GUI-generated viewers, but a simpler fix might be to pass `showValidations = False` to `createViewer` to disable this entirely in the GUI.  As far as I can tell there isn't a great way to get the validation messages associated with an individual report in Arelle; we're forced to grab the validation messages from the buffer from the appropriate period, which works OK from the command line when all we're doing is opening the document, but is a bit trickier in the GUI when the user may do other things that add messages to the buffer.

iXBRLViewerPlugin/iXBRLViewer.py
sagesmith-wf: ### tl;dr 
1. It is impossible to actually save the viewer with the current flow
2. the save viewer lets you change which version of the js file

### long winded explaination
If you have the plugin loaded in arelle and you are using the GUI the ixbrlViewer will automatically generate and open itself once you load a model [1](https://github.com/Workiva/ixbrl-viewer/blob/7bcde691a5edd59487e9e3158554dffb14c7886c/iXBRLViewerPlugin/__init__.py#L152), [2](https://github.com/Workiva/ixbrl-viewer/blob/7bcde691a5edd59487e9e3158554dffb14c7886c/iXBRLViewerPlugin/__init__.py#L136-L139). The viewer that it generates is using whatever `ixbrlviewer.js` is packaged with the plugin so in my case I am using the one I locally built `iXBRLVIewerPlugin/viewer/dist/ixbrlviewer.js`.  When you hit the save ixbrl viewer option in the arelle gui at this point (see the issue for screenshots) it will ask you for a link to the `ixbrlviewer.js` that you want to save using (at this point I switch to the [CDN server version](https://github.com/Workiva/ixbrl-viewer#accessing-via-the-cdn)). The viewer is only parsing the xmlDocument once and then tries to modify the same model (maybe a mistake?) when you try to hit save, but because it already generated a viewer prior to you hitting save (on document load) then it is currently stuck always returning `False` [here](https://github.com/Workiva/ixbrl-viewer/blob/7bcde691a5edd59487e9e3158554dffb14c7886c/iXBRLViewerPlugin/iXBRLViewer.py#L303) which then makes it skip saving [1](https://github.com/Workiva/ixbrl-viewer/blob/7bcde691a5edd59487e9e3158554dffb14c7886c/iXBRLViewerPlugin/iXBRLViewer.py#L368-L369), [2](https://github.com/Workiva/ixbrl-viewer/blob/7bcde691a5edd59487e9e3158554dffb14c7886c/iXBRLViewerPlugin/__init__.py#L100-L102)

iXBRLViewerPlugin/__init__.py
sagesmith-wf: > `showValidations = False` to `createViewer`

I'm a lot happier with that suggestion it surcumvents all of the complexity of the problem 👍 , I will make it so

aaroncameron-wk (dismissed):

iXBRLViewerPlugin/iXBRLViewer.py
brettkail-wk: It seems wrong to reopen the already-ixbrl-viewer-enhanced XHTML file rather than reopening the source XHTML file when adding a different ixbrlviewer.js.  It seems like we should fix that rather than changing this error behavior.

austinmatherne-wk (approved):

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -351,17 +352,19 @@ def createViewer(self, scriptUrl="js/dist/ixbrlviewer.js", showValidations = Tru
...
-                iv.addFile(iXBRLViewerFile(docSetFiles[n], docSet[n].xmlDocument))
+            for filename, xmlDocument in xmlDocsByFilename.items():
+                iv.addFile(iXBRLViewerFile(filename, xmlDocument))
 
-            xmlDocument = docSet[0].xmlDocument 
+            xmlDocument = next(iter(xmlDocsByFilename.values()))
 
         else:
-            xmlDocument = dts.modelDocument.xmlDocument
+            xmlDocument = deepcopy(dts.modelDocument.xmlDocument)
```
paulwarren-wk, line 367: What are these changes for?

sagesmith-wf: Hopefully the comment opens since its resolved https://github.com/Workiva/ixbrl-viewer/pull/371#discussion_r1095187534 but since the plugin opens on startup it modifies `dts.modelDocument.xmlDocument` to have the ixbrlviewer json and javascript added. Then when you go to save there is a check to make sure you aren't modifying a document with an ixbrl-viewer (which you always are) and thus prevents you from saving. The deepcopy lets you create an ixbrl-viewer as many times as you want from the same document without ever modifying the original source document which is important since the save functionality allows you to supply a URL

paulwarren-wk: Ah yes.  That makes sense, thank you.

paulwarren-wk (approved):

aaroncameron-wk (approved):

derekgengenbacher-wf (approved):

aaroncameron-wk:
QA +10
Saved as expected locally.

rmconsole-wf (approved):
+1 from RM

PR #384, opened 2023-02-14
Prevent CSS from being incorrectly escaped

FredrikGeers:
#### Reason for change
By writing out the html with an xml method, the characters in the entire document are escaped, including content in `<script>` and `<style>` tags. In style tags css the child combinator (`>`) can be used, which is not functioning anymore after it is incorrectly escaped to `&gt;`.

#### Description of change
Changed the lxml.etree toString method parameter to write out using the html method.

#### Steps to Test
Added automated testcase for this situation.
Test manually by adding viewer to ixbrl report that uses > operator in the css in a style tag. Expected result is that this operator remains untouched.

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/xhtmlserialize.py
+++ b/iXBRLViewerPlugin/xhtmlserialize.py
@@ -38,4 +38,4 @@ class XHTMLSerializer:
 
     def serialize(self, xmlDocument, fout):
         self._expandEmptyTags(xmlDocument)
-        fout.write(etree.tostring(xmlDocument, method="xml", encoding="utf-8", xml_declaration=True))
+        fout.write(etree.tostring(xmlDocument, method="html", encoding="utf-8"))
--- a/tests/unit_tests/iXBRLViewerPlugin/test_xhtmlserialize.py
+++ b/tests/unit_tests/iXBRLViewerPlugin/test_xhtmlserialize.py
@@ -14,6 +14,16 @@ class TestXHTMLSerializer(unittest.TestCase):
     def _html(self, s):
         return '<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body>%s</body></html>' % s
 
+    def _htmlHead(self, s):
+        return '<html xmlns="http://www.w3.org/1999/xhtml"><head>%s</head><body></body></html>' % s   
+
+    def _serialize(self, html):
+        doc = lxml.etree.fromstring(html)
+        f = io.BytesIO()
+        writer = XHTMLSerializer()
+        writer.serialize(doc, f)
+        return f.getvalue().decode('utf-8')     
+
     def _checkTagExpansion(self, html_in, html_out):
         writer = XHTMLSerializer()
         doc = lxml.etree.fromstring(self._html(html_in))
@@ -41,11 +51,11 @@ class TestXHTMLSerializer(unittest.TestCase):
 
     def test_serialize(self):
         htmlsrc = self._html("<p>hello</p>")
-        doc = lxml.etree.fromstring(htmlsrc)
-        f = io.BytesIO()
+        serialized = self._serialize(htmlsrc)
+        self.assertEqual(serialized, htmlsrc)
 
-        writer = XHTMLSerializer()
-        writer.serialize(doc, f)
-
-        # XML declaration should be added.
-        self.assertEqual(f.getvalue().decode('utf-8'), "<?xml version='1.0' encoding='utf-8'?>\n" + htmlsrc)
+    def test_serializeCss(self):
+        # Do not escape > character in CSS style tag
+        htmlsrc = self._html('<style type="text/css">li > p { color: red; }</style>')
+        serialized = self._serialize(htmlsrc)
+        self.assertEqual(serialized, htmlsrc)


aviary-wf:
## Security Insights
<details>
<li>Watched keyword <code>etree</code> in <code>iXBRLViewerPlugin/xhtmlserialize.py</code> line(s) ['41'] added</li>
</details>

## Action Items
---

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -38,4 +38,4 @@ def _expandEmptyTags(self, xml):
 
     def serialize(self, xmlDocument, fout):
         self._expandEmptyTags(xmlDocument)
-        fout.write(etree.tostring(xmlDocument, method="xml", encoding="utf-8", xml_declaration=True))
+        fout.write(etree.tostring(xmlDocument, method="html", encoding="utf-8"))
```
paulwarren-wk, line 41: I'm pretty sure that this will make the document XML invalid, which means that they won't open if treated as XHTML, which they often will be as the `.xhtml` extension is in common use.

We've tried to make it so that the viewer output is valid as both HTML and XHTML, but annoyingly, HTML and XHTML have incompatible escaping rules for `<script>` and `<style>` tags.  This hasn't been a problem with SEC filings, as SEC requires all styling to be done on style attributes, but ESEF permits the use of `<style>` elements.

Possible solutions:

1. I think that the only problem character in CSS is `>`, and escaping that is optional in XML, so one solution would be to post-process the output and unescape the`>` (or write a custom serialiser).

2. We may be able to use `<![CDATA[` in a CSS comment to wrap the CSS and then not escape the CSS content.  

3. Give up trying to be both XHTML and HTML compatible at the same time.  I'm reluctant to do this, as the file extension is only one factor in how browsers choose to treat a document.

FredrikGeers: Using a style tag with this character in it does not make the document invalid xhtml, and any browser will be able to process the result properly. Perhaps there can be cases where generic xml parsing of the result will cause issues. I would argue though that the ixbrl-viewer result is meant for consumption in a browser, and generic parsing of an ixbrl report should be done on the report itself instead.

Responding to your suggested solutions:

1. This is not limited to the `>` character, as there is also an `&` operator in css which could have the same problem. That does not mean that this proposed solution is not an option though.
2. I have tried this before suggesting a fix in this project, but this seemed not to have the expected results.
3. I do not think this is necessary. I have now dropped the xml declaration in the results because lxml does not offer that option for the html method, but we could add this back in if you believe that is needed.

paulwarren-wk: > Using a style tag with this character in it does not make the document invalid xhtml

No, but not using the XML serializer risks breaking it in other ways.  For example, the [lxml docs](https://lxml.de/tutorial.html#serialisation) show this example:

```python
>>> etree.tostring(root, method='xml') # same as above
b'<html><head/><body><p>Hello<br/>World</p></body></html>'

>>> etree.tostring(root, method='html')
b'<html><head></head><body><p>Hello<br>World</p></body></html>'
```

Clearly the `<br>` tag makes the latter invalid XML, and thus invalid XHTML and I get an error message if I try to open it as XHTML in a browser.

Interestingly, if the document is in the XHTML namespace, which it will be for iXBRL documents, `lxml` appears to not remove the closing `<br>` tag, but instead expands it to `<br></br>`, which results in a double line break if the document is rendered in HTML mode.

```python
>>> root = etree.XML('<html xmlns="http://www.w3.org/1999/xhtml"><head/><body><p>Hello<br />World</p></body></html>')
>>> etree.tostring(root, method='html')
b'<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body><p>Hello<br></br>World</p></body></html>'
```

Cross linking https://github.com/Workiva/ixbrl-viewer/pull/379#discussion_r1102073372 

OOI, what is the `&` in CSS for?  I'm only aware of it in SASS/Less.

FredrikGeers: You are correct, this `&` is not available in pure CSS, my mistake.
The `<br>` example is a good reason for not accepting this PR in the current form indeed.

Perhaps making your option 2 work is the cleanest solution, so then it's worth an investigation on why that currently isn't the case.

paulwarren-wk: Combined with the issues on #379, I think we need a custom XHTML/HTML serializer.  If we had that, we could either have it not escape `>` (either universally, or just within `<style>` elements), or use the CDATA approach – I don't think either approach is possible using lxml's existing serializers.  I'll look into implementing this.

Another edge case is the content of `content` properties, e.g.:

```css
div:before { content: "<foo>"; }
div:before { content: "&lt;foo&gt;"; }
div:before { content: "\u00003cfoo>"; }
```

The first is invalid XHTML, the second renders differently in HTML and XHTML, the third one renders consistently in both.

FredrikGeers:
Closing this proposal since #386 is accomplishing the same, but in a better way.

PR #386, opened 2023-02-20
XT-2790 Fix escaping of style element contents (add custom XHTML serializer)

paulwarren-wk:
#### Reason for change

At present, the XHTML produced by the viewer builder escapes `<`, `&` and `>` throughout the document.  This is fine as long as the result is treated as XHTML.   Ordinary HTML expects different escaping to `<style>` and `<script>` elements.  In particular, an escaped `>` in a style element will not be recognised as valid CSS, resulting in different rendering in HTML and XHTML rendering mode.

#### Description of change

With this PR, the viewer does not escape `>` appearing within `<style>` elements (this is optional in XML/XHTML).

`<` and `&` are escaped as if they are appearing within a CSS string using `\NNNNNN` hex escapes (this is the only place other than comments in which they can legally appear in CSS).

This is an alternative solution for #384 .

#### Steps to Test

Create an iXBRL document with an inline style sheet, and a style rule such as:

```css
p > span { color: #f00 }
```

Create a viewer from it, and confirm that the `>` is not escaped in the resulting output.

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -396,8 +396,8 @@ class iXBRLViewer:
                 for f in self.files:
                     self.dts.info("viewer:info", "Saving in output zip %s" % f.filename)
                     fout = attrdict(write=lambda s: zout.writestr(_outPrefix + f.filename, s))
-                    writer = XHTMLSerializer()
-                    writer.serialize(f.xmlDocument, fout)
+                    writer = XHTMLSerializer(fout)
+                    writer.serialize(f.xmlDocument)
                 zout.write(os.path.join(os.path.dirname(__file__), "viewer", "dist", "ixbrlviewer.js"), _outPrefix + "ixbrlviewer.js")
         elif os.path.isdir(outPath):
             # If output is a directory, write each file in the doc set to that
@@ -406,8 +406,8 @@ class iXBRLViewer:
                 filename = os.path.join(outPath, "{0[0]}{1}{0[1]}".format(os.path.splitext(f.filename), outBasenameSuffix))
                 self.dts.info("viewer:info", "Writing %s" % filename)
                 with open(filename, "wb") as fout:
-                    writer = XHTMLSerializer()
-                    writer.serialize(f.xmlDocument, fout)
+                    writer = XHTMLSerializer(fout)
+                    writer.serialize(f.xmlDocument)
 
         else:
             if len(self.files) > 1:
@@ -421,5 +421,5 @@ class iXBRLViewer:
             else:
                 self.dts.info("viewer:info", "Writing %s" % outPath)
                 with open("{0[0]}{1}{0[1]}".format(os.path.splitext(outPath), outBasenameSuffix), "wb") as fout:
-                    writer = XHTMLSerializer()
-                    writer.serialize(self.files[0].xmlDocument, fout)
+                    writer = XHTMLSerializer(fout)
+                    writer.serialize(self.files[0].xmlDocument)
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -395,9 +395,9 @@ class iXBRLViewer:
             with zipfile.ZipFile(outPath, "a", zipfile.ZIP_DEFLATED, True) as zout:
                 for f in self.files:
                     self.dts.info("viewer:info", "Saving in output zip %s" % f.filename)
-                    fout = attrdict(write=lambda s: zout.writestr(_outPrefix + f.filename, s))
-                    writer = XHTMLSerializer(fout)
-                    writer.serialize(f.xmlDocument)
+                    with zout.open(_outPrefix + f.filename, "w") as fout:
+                        writer = XHTMLSerializer(fout)
+                        writer.serialize(f.xmlDocument)
                 zout.write(os.path.join(os.path.dirname(__file__), "viewer", "dist", "ixbrlviewer.js"), _outPrefix + "ixbrlviewer.js")
         elif os.path.isdir(outPath):
             # If output is a directory, write each file in the doc set to that
--- a/iXBRLViewerPlugin/xhtmlserialize.py
+++ b/iXBRLViewerPlugin/xhtmlserialize.py
@@ -92,8 +92,10 @@ class XHTMLSerializer:
         if s is None:
             return ''
         if escape_mode == EscapeMode.STYLE:
-            # Don't escape >
-            # & and < may only appear in a string or comment in CSS, so escape
+            # HTML does not understand XML escapes inside a style element.
+            # Don't escape ">".  Escaping isn't required by XML, but is common
+            # practice, so we do it elsewhere.
+            # "&" and "<" may only appear in a string or comment in CSS, so escape
             # as if they're in a string.
             return re.sub(r'([<&])', lambda m: "\\%06X" % ord(m.group(1)), s)
         return re.sub(r'([<>&])', lambda m: self.ENTITIES[m.group(0)], s)
--- a/iXBRLViewerPlugin/xhtmlserialize.py
+++ b/iXBRLViewerPlugin/xhtmlserialize.py
@@ -110,8 +110,9 @@ class XHTMLSerializer:
             # HTML does not understand XML escapes inside a style element.
             # Don't escape ">".  Escaping isn't required by XML, but is common
             # practice, so we do it elsewhere.
-            # "&" and "<" may only appear in a string or comment in CSS, so escape
-            # as if they're in a string.
+            # "&" and "<" are escaped using XML escapes.  This will break
+            # HTML/XHTML compatibility, but any source document using them
+            # wouldn't have been HTML compatible anyway.
             self.write(self.STYLE_ESCAPE_RE.sub(lambda m: self.escape_str(m[0]),s))
         else:
             self.write(self.ESCAPE_RE.sub(lambda m: self.escape_str(m[0]),s))
--- a/iXBRLViewerPlugin/xhtmlserialize.py
+++ b/iXBRLViewerPlugin/xhtmlserialize.py
@@ -17,6 +17,7 @@ from enum import Enum
 import re
 
 XHTML_NS = 'http://www.w3.org/1999/xhtml'
+XML_NS = 'http://www.w3.org/XML/1998/namespace'
 
 class EscapeMode(Enum):
     DEFAULT = 0
@@ -65,7 +66,10 @@ class XHTMLSerializer:
         qname = etree.QName(tag)
         if qname.namespace is None:
             return qname.localname
-        prefix = next(iter(sorted((p for p, ns in nsmap.items() if ns == qname.namespace and p is not None), key = self.prefix_sort)))
+        if qname.namespace == XML_NS:
+            prefix = 'xml'
+        else:
+            prefix = next(iter(sorted((p for p, ns in nsmap.items() if ns == qname.namespace and p is not None), key = self.prefix_sort)))
         return "%s:%s" % (prefix, qname.localname)
 
     def is_selfclosable(self, n):
--- a/iXBRLViewerPlugin/xhtmlserialize.py
+++ b/iXBRLViewerPlugin/xhtmlserialize.py
@@ -37,24 +37,36 @@ class XHTMLSerializer:
         '&': '&amp;'
         }
 
-    def __init__(self, xml_declaration = True, assume_xhtml = True):
+    def __init__(self, fout, xml_declaration = True, assume_xhtml = True):
+        self.fout = fout
+        self.encoding = "utf-8"
         self.xml_declaration = xml_declaration
         self.assume_xhtml = assume_xhtml
+
+    def write(self, s):
+        self.fout.write(s.encode(self.encoding))
 
     def prefix_sort(self, p):
         return p if p is not None else '0'
 
-    def tag_as_qname(self, tag, nsmap):
+    def qname_for_node(self, node):
+        qname = etree.QName(node.tag)
+        if qname.namespace is None:
+            prefix = None
+        elif node.nsmap.get(node.prefix, None) == qname.namespace:
+            prefix = node.prefix
+        else:
+            prefix = next(iter(sorted((p for p, ns in nsmap.items() if ns == qname.namespace), key = self.prefix_sort)))
+        if prefix is None:
+            return qname.localname
+        return "%s:%s" % (prefix, qname.localname)
+
+    def qname_for_attr(self, tag, nsmap):
         qname = etree.QName(tag)
         if qname.namespace is None:
             return qname.localname
-
-        prefix = next(iter(sorted((p for p, ns in nsmap.items() if ns == qname.namespace), key = self.prefix_sort)))
-
-        if prefix is None:
-            return qname.localname
-        else:
-            return "%s:%s" % (prefix, qname.localname)
+        prefix = next(iter(sorted((p for p, ns in nsmap.items() if ns == qname.namespace and p is not None), key = self.prefix_sort)))
+        return "%s:%s" % (prefix, qname.localname)
 
     def is_selfclosable(self, n):
         qname = etree.QName(n)
@@ -88,22 +100,22 @@ class XHTMLSerializer:
 
     def attributes(self, node):
         return sorted(
-            '%s="%s"' % (self.tag_as_qname(k, node.nsmap), self.escape_attr(v))
+            '%s="%s"' % (self.qname_for_attr(k, node.nsmap), self.escape_attr(v))
                 for k, v in node.items()
             )
 
     def write_comment(self, n, escape_mode):
-        return '<!--' + n.text + '-->' + self.escape_text(n.tail, escape_mode)
+        self.write('<!--' + n.text + '-->' + self.escape_text(n.tail, escape_mode))
 
     def write_processing_instruction(self, n, escape_mode):
-        return '<?' + n.target + ' ' + n.text + '?>' + self.escape_text(n.tail, escape_mode)
+        self.write( '<?' + n.target + ' ' + n.text + '?>' + self.escape_text(n.tail, escape_mode))
 
     def write_element(self, n, parent_nsmap = {}, escape_mode = EscapeMode.DEFAULT):
-        name = self.tag_as_qname(n.tag, n.nsmap)
+        name = self.qname_for_node(n)
         qname = etree.QName(n.tag)
         selfclose = len(n) == 0 and n.text is None and self.is_selfclosable(n)
         parts = [ name ] + self.namespace_declarations(n.nsmap, parent_nsmap) + self.attributes(n)
-        s = '<' + ' '.join(parts)
+        self.write('<' + ' '.join(parts))
 
         if qname.localname.lower() == 'style':
             inner_escape_mode = EscapeMode.STYLE
@@ -111,27 +123,24 @@ class XHTMLSerializer:
             inner_escape_mode = escape_mode
 
         if selfclose:
-            s += '/>'
+            self.write('/>')
         else:
-            s += '>'
-            s += self.escape_text(n.text, inner_escape_mode)
+            self.write('>' + self.escape_text(n.text, inner_escape_mode))
 
             for child in n.iterchildren():
                 if isinstance(child, etree._Comment):
-                    s += self.write_comment(child, inner_escape_mode)
+                    self.write_comment(child, inner_escape_mode)
                 elif isinstance(child, etree._ProcessingInstruction):
-                    s += self.write_processing_instruction(child, inner_escape_mode)
+                    self.write_processing_instruction(child, inner_escape_mode)
                 else:
-                    s += self.write_element(child, n.nsmap, inner_escape_mode)
-            s += '</%s>' % name
+                    self.write_element(child, n.nsmap, inner_escape_mode)
+            self.write('</%s>' % name)
 
-        s += self.escape_text(n.tail, escape_mode)
+        self.write(self.escape_text(n.tail, escape_mode))
 
-        return s
-
-    def serialize(self, element, fout):
+    def serialize(self, element):
         if hasattr(element, 'getroot'):
             element = element.getroot()
         if self.xml_declaration:
-            fout.write('<?xml version="1.0" encoding="utf-8"?>\n'.encode('utf-8'))
-        fout.write(self.write_element(element).encode('utf-8'))
+            self.write('<?xml version="1.0" encoding="utf-8"?>\n')
+        self.write_element(element)
--- a/iXBRLViewerPlugin/xhtmlserialize.py
+++ b/iXBRLViewerPlugin/xhtmlserialize.py
@@ -151,10 +151,20 @@ class XHTMLSerializer:
 
         self.write_escape_text(n.tail, escape_mode)
 
-    def serialize(self, element):
+    def write_xml_declaration(self, docinfo = None):
         if self.xml_declaration:
-            self.write('<?xml version="1.0" encoding="utf-8"?>\n')
+            version = "1.0"
+            standalone = ""
+            if docinfo is not None:
+                version = docinfo.xml_version
+                if docinfo.standalone:
+                    standalone = ' standalone="yes"'
+            self.write('<?xml version="%s" encoding="%s"%s?>\n' % (version, self.encoding, standalone))
+
+    def serialize(self, element):
         if hasattr(element, 'getroot'):
+            self.write_xml_declaration(element.docinfo)
+
             element = element.getroot()
             while element.getprevious() is not None:
                 element = element.getprevious()
@@ -166,5 +176,6 @@ class XHTMLSerializer:
                 element = element.getnext()
 
         else:
+            self.write_xml_declaration()
             self.write_element(element)
 
--- a/iXBRLViewerPlugin/xhtmlserialize.py
+++ b/iXBRLViewerPlugin/xhtmlserialize.py
@@ -93,7 +93,7 @@ class XHTMLSerializer:
         return sorted(self.xmlns_declaration(p, new_nsmap[p]) for p in changed)
 
     def escape_char(self, c):
-        return self.ENTITIES.get(c, '&#x%2X;' % ord(c))
+        return self.ENTITIES.get(c, '&#x%02X;' % ord(c))
 
     def write_escape_text(self, s, escape_mode):
         if s is None:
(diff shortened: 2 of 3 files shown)


aviary2-wf:
## Security Insights
<details>
<li>Watched keyword <code>etree</code> in <code>iXBRLViewerPlugin/xhtmlserialize.py</code> line(s) ['48', '60', '103', '120', '122'] added</li>
</details>

## Action Items
---

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,125 @@
...
+
+    def prefix_sort(self, p):
+        return p if p is not None else '0'
+
+    def tag_as_qname(self, tag, nsmap):
+        qname = etree.QName(tag)
+        if qname.namespace is None:
+            return qname.localname
+
+        prefix = next(iter(sorted((p for p, ns in nsmap.items() if ns == qname.namespace), key = self.prefix_sort)))
```
brettkail-wk, line 52: What is the intent of this `sorted`?  If it's not possible to determine the prefix that was actually used(?), I would think we always want the most recently bound prefix.

I guess it won't matter much in practice since most documents will predefine all their namespaces on the root anyway.  I wonder if it's worth maintaining a ns-to-prefix map rather than iterating all?

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,125 @@
...
+            for child in n.iterchildren():
+                if isinstance(child, etree._Comment):
+                    s += self.write_comment(child, inner_escape_mode)
+                elif isinstance(child, etree._ProcessingInstruction):
+                    s += self.write_processing_instruction(child, inner_escape_mode)
+                else:
+                    s += self.write_element(child, n.nsmap, inner_escape_mode)
+            s += '</%s>' % name
+
+        s += self.escape_text(n.tail, escape_mode)
```
brettkail-wk, line 128: I am worried that string concatenation will perform poorly for very large files, so I think we need to find some way to stream all the output rather than using string concatenation at all.

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,125 @@
...
+
+        return s
+
+    def serialize(self, element, fout):
+        if hasattr(element, 'getroot'):
+            element = element.getroot()
```
brettkail-wk, line 134: Is it possible to preserve comments outside the root element?

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,125 @@
...
+        return '<?' + n.target + ' ' + n.text + '?>' + self.escape_text(n.tail, escape_mode)
+
+    def write_element(self, n, parent_nsmap = {}, escape_mode = EscapeMode.DEFAULT):
+        name = self.tag_as_qname(n.tag, n.nsmap)
+        qname = etree.QName(n.tag)
+        selfclose = len(n) == 0 and n.text is None and self.is_selfclosable(n)
+        parts = [ name ] + self.namespace_declarations(n.nsmap, parent_nsmap) + self.attributes(n)
+        s = '<' + ' '.join(parts)
+
+        if qname.localname.lower() == 'style':
```
brettkail-wk, line 108: I think this `lower()` is not needed for valid XHTML?

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,125 @@
...
+
+    def escape_text(self, s, escape_mode):
+        if s is None:
+            return ''
+        if escape_mode == EscapeMode.STYLE:
+            # Don't escape >
+            # & and < may only appear in a string or comment in CSS, so escape
+            # as if they're in a string.
+            return re.sub(r'([<&])', lambda m: "\\%06X" % ord(m.group(1)), s)
+        return re.sub(r'([<>&])', lambda m: self.ENTITIES[m.group(0)], s)
```
brettkail-wk, line 87: I guess escaping `>` is never required?  You might add a comment to the "Don't escape" above.

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,125 @@
...
+
+        s += self.escape_text(n.tail, escape_mode)
+
+        return s
+
+    def serialize(self, element, fout):
+        if hasattr(element, 'getroot'):
+            element = element.getroot()
+        if self.xml_declaration:
+            fout.write('<?xml version="1.0" encoding="utf-8"?>\n'.encode('utf-8'))
```
brettkail-wk, line 136: Out of curiosity, does iXBRL tolerate 1.1?

Can/should we include `standalone="..."` if present in the source?

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: The `sorted` is to ensure that the output documents are stable and predictable (for testing purposes) and to ensure that the default prefix is preferred over an explicit prefix.

I don't think we can get access to the prefix used in the source.

Maintaining an NS to prefix map would be reasonable, as we already detect when the prefix bindings have changed, although keeping track of the most recently bound would be a bit awkward, and wouldn't help with the situation where both prefixes are bound at the same time.

brettkail-wk: Got it, thanks.

For elements, it looks like `n.prefix` works.  For attributes, I also don't see a way to get the original prefix from lxml; how unfortunate.  Anyway, duplicate prefixes are hopefully rare.

paulwarren-wk: Ah, not sure how I missed `n.prefix`.  I'll update to use that, assuming it's in scope and bound to the right URI.  Which it should be given that we don't support editing of the document.

For attributes, I realise that there's an issue with the current code which we mustn't use an unprefixed attribute name if the attribute has a namespace, because the default namespace binding doesn't apply to attributes.  I think one of the places where duplicate NS bindings do get used is to provide a default binding for elements, and an explicit binding for attributes (and QName-valued elements).

paulwarren-wk: Done in 04c08600ec7621436fcf0e5665e283e23976d94c

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Done in 04c08600ec7621436fcf0e5665e283e23976d94c

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Removed in 3e73f99

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -396,8 +396,8 @@ def save(self, outPath, outBasenameSuffix="", outzipFilePrefix=""):
                 for f in self.files:
                     self.dts.info("viewer:info", "Saving in output zip %s" % f.filename)
                     fout = attrdict(write=lambda s: zout.writestr(_outPrefix + f.filename, s))
-                    writer = XHTMLSerializer()
-                    writer.serialize(f.xmlDocument, fout)
+                    writer = XHTMLSerializer(fout)
```
brettkail-wk, line 399: I don't know Python well, but do we need some kind of output stream buffering, or does ZipFile do that for us?

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,134 @@
...
+
+    def write_processing_instruction(self, n, escape_mode):
+        self.write( '<?' + n.target + ' ' + n.text + '?>' + self.escape_text(n.tail, escape_mode))
+
+    def write_element(self, n, parent_nsmap = {}, escape_mode = EscapeMode.DEFAULT):
+        name = self.qname_for_node(n)
+        qname = etree.QName(n.tag)
+        selfclose = len(n) == 0 and n.text is None and self.is_selfclosable(n)
+        parts = [ name ] + self.namespace_declarations(n.nsmap, parent_nsmap) + self.attributes(n)
+        self.write('<' + ' '.join(parts))
```
brettkail-wk, line 120: It might be worth streaming the attributes individually since we've seen large `` and might potentially have class/style/etc. attributes also applied.

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,134 @@
...
+
+        if qname.localname == 'style':
+            inner_escape_mode = EscapeMode.STYLE
+        else:
+            inner_escape_mode = escape_mode
+
+        if selfclose:
+            self.write('/>')
+        else:
+            self.write('>' + self.escape_text(n.text, inner_escape_mode))
```
brettkail-wk, line 130: It might be worth streaming the `escape_text` call too: we have seen enormous `<style>` with embedded fonts, background images, etc.  It might even be nice to avoid the `.replace` call and manually splitting the string and writing it out in chunks.

@derekgengenbacher-wf It's probably worth testing this on some very large documents (perhaps SEC, ESEF, and FERC?) to verify that this doesn't cause any perf problems with time/memory or any rendering issues.

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Comment expanded.

brettkail-wk (dismissed):

iXBRLViewerPlugin/iXBRLViewer.py
paulwarren-wk: Yes, I think you're right - this is broken.  The above code is relying on `fout` being written to in a single call to `write()` , which is not the norm for a Python "file-like object".

I think this needs to be rewritten to use `BytesIO`.

paulwarren-wk: This is now done in 52b1586773ea386ac8764f026dfd4b33d1ee08d8

tests/unit_tests/iXBRLViewerPlugin/test_xhtmlserialize.py
```diff
@@ -140,6 +140,11 @@ def test_custom_xml_serializer(self):
                 r'''<div attr1='"foo"'>abc</div>''',
                 r'''<div attr1="&quot;foo&quot;">abc</div>''',
             ),
+            # xml prefix doesn't need declaring
+            (
+                r'''<div xml:lang="en" attr='abc'>abc</div>''',
```
sagesmith-wf, line 145: Should `attr='abc'` be `attr="abc"`? Above testcase implies maybe to me? Also why `r` this doesn't look like a regex?

sagesmith-wf (approved):

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Done in 1fd8e955af60e8d24e9056053042d931674e5702

I've left the `re.sub` in.  Testing with a fairly large ESEF document, it adds around 0.2s to an overall write time of 5s compared to writing the string out unmodified, so that's the upper bound of what we could save here.

Naive attempts to reimplement it in a streaming manner are significantly slower.  `re.sub` is implemented in C, so I think it may be hard to improve on in Python.

tests/unit_tests/iXBRLViewerPlugin/test_xhtmlserialize.py
paulwarren-wk: > Should `attr='abc'` be `attr="abc"`

This was copy and paste from the previous test, which is checking that attribute quotes are normalized to double quotes, regardless of contents, but I think it's harmless here, as it's just an equivalent input document.

>  Also why `r` this doesn't look like a regex?

`r` is for any raw string, and avoids having to escape `\`.  It's only actually needed for the two tests that the `style` escaping to unicode escape sequences, but it's harmless to use it for all tests.  Similarly, we only need the `'''` for the tests that mix `'` and `"`, but using it consistently saves having to think and makes the strings line up nicely :)

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Done in https://github.com/Workiva/ixbrl-viewer/commit/1fd8e955af60e8d24e9056053042d931674e5702

sagesmith-wf (dismissed):

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Just about - see fdc52a1.

You can get at the comment/PI nodes, but whitespace between them does not appear to be captured, so the above code just blindly adds a single new line between each one.

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: > Out of curiosity, does iXBRL tolerate 1.1?

I don't recall any discussion of this, and the iXBRL spec references XML 1.0, so I think probably not.  That said, it's easy enough to make this code parrot the version number of the input document.  

It doesn't look like lxml supports 1.1, in that it still rejects characters that are now allowed, but I've implemented the new 1.1 requirement that certain control characters are escaped.

> Can/should we include standalone="..." if present in the source?

I see no reason not to repeat include it if the source document specifies it, so I've made it do that.

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,172 @@
...
+        if prefix is None:
+            return 'xmlns="%s"' % self.escape_attr(uri)
+        return 'xmlns:%s="%s"' % (prefix, self.escape_attr(uri))
+
+    def namespace_declarations(self, new_nsmap, cur_nsmap):
+        changed = set(p[0] for p in (set(new_nsmap.items()) ^ set(cur_nsmap.items())))
+        return sorted(self.xmlns_declaration(p, new_nsmap[p]) for p in changed)
+
+    def escape_char(self, c):
+        return self.ENTITIES.get(c, '&#x%2X;' % ord(c))
```
brettkail-wk, line 96: I think remove the padding:
```
$ echo $'<r>\n</r>' | python b.py | xmllint -
-:2: parser error : CharRef: invalid hexadecimal value
<r>&#x A;</r>
      ^
```
Well, it might be worth writing out at least \t\r\n directly.

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,172 @@
...
+    def write_escape_text(self, s, escape_mode):
+        if s is None:
+            return
+        if escape_mode == EscapeMode.STYLE:
+            # HTML does not understand XML escapes inside a style element.
+            # Don't escape ">".  Escaping isn't required by XML, but is common
+            # practice, so we do it elsewhere.
+            # "&" and "<" may only appear in a string or comment in CSS, so escape
+            # as if they're in a string.
+            self.write(re.sub(r'([<&])', lambda m: "\\%06X" % ord(m.group(1)), s))
```
brettkail-wk, line 107: > `<` and `&` are escaped as if they are appearing within a CSS string using `\NNNNNN` hex escapes (this is the only place other than comments in which they can legally appear in CSS).

I am somewhat nervous this will not remain true (`&` in [CSS Nesting Module working draft](https://www.w3.org/TR/css-nesting-1/)); files using that are obviously not in the XHTML/HTML subset, but it seems wrong to break the XHTML styling.  Well, it feels a little off to write `\` escapes in a non-string context.

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,172 @@
...
+            self.write(' %s="' % qname)
+            self.write(self.escape_attr(value))
+            self.write("\"")
+
+    def write_comment(self, n, escape_mode):
+        self.write('<!--' + n.text + '-->')
+        self.write_escape_text(n.tail, escape_mode)
+
+    def write_processing_instruction(self, n, escape_mode):
+        self.write( '<?' + n.target + ' ' + n.text + '?>')
```
brettkail-wk, line 122: It's odd that `<?pi?>` round-trips as `<?pi ?>`.  It might be possible to distinguish the two(?), but the former seems preferable.  I somehow doubt PIs are in regular use anyway :-).

iXBRLViewerPlugin/xhtmlserialize.py
brettkail-wk, line 107: FWIW, I remembered that `]]>` is invalid.  I doubt there's any use-case for that (even in a `content: ']]&gt;'` string).  If you'd rather not try to fix that, a comment saying we're not trying to handle it seems fine.

brettkail-wk (commented):
Overall looking good.

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Ah, I did wonder if CSS syntax might start using them at some point.  If it does, I think we're stuck for maintaining XHTML/HTML compatibility.

For now, I think the options are:

* CSS parse the contents to find out if we're in a string (and either raise an error, or warn if not)
* Don't escape them at all (and raise an error/warn if they occur)

paulwarren-wk: Oh, I thought `]]>` was only significant once you were inside a `CDATA` section, but you're right.

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Oh dear, this commit was a mess!  I was aiming for `%02X` so that it's zero padded, and I misread the bit about `0x01` to `0x1F` being newly allowed in 1.1 as them _all_ being previously forbidden in 1.0, when obviously that can't be right.  And all the multi-line tests that we have so far neatly dodge this code because they're only testing normalisation of whitespace outside the root element.

iXBRLViewerPlugin/xhtmlserialize.py
brettkail-wk: If a valid iXBRL input document contains `<style>div::before { content: '&amp;&lt;'; }</style>`, then the producer _already_ did not care about HTML compatibility or else they would have used `'\26\3c'`.  ixbrl-viewer currently does not make this worse (the output document is as compatible with HTML as the input document), and I don't think we need any code to improve that, so we can just use `escape_char`.

However, if a valid iXBRL input document uses `<style>div > span { color: red; }</style>`, then the iXBRL producer might have been trying to care about HTML compatibility.  ixbrl-viewer currently breaks that by outputting `&gt;`, and I am interested in fixing this case because we have third party iXBRL producers that output styles like that.  I think a regex like `r'([<&]|]]>)'` could be used to fix those cases, including the extreme edge case of `<style>div::before { content: ']]&gt;' }</style>`.

I think the PR is currently missing handling of control character escapes in `<style>` for XML 1.0.  I don't know if there's a use case for a literal control character in a string in `<style>`, but I guess we could conditionally escape the control characters based on whether the XML version is 1.0.

To summarize, I suggest:
- Remove the special handling of `style` altogether.
- Add `]]>` to the regex.  (I guess `escape_char`/`ENTITIES` will be weirdly named, so perhaps remove the function altogether and just use an `ESCAPES` dict directly.)
- In addition to the escape padding fix and \t\r\n improvement discussion, consider only conditionally escaping control characters based on 1.0.

brettkail-wk: I gave a suggestion for fixing this [in the other thread](https://github.com/Workiva/ixbrl-viewer/pull/386#discussion_r1116903284).

paulwarren-wk: > * Remove the special handling of `style` altogether.
> * Add `]]>` to the regex. 

To be clear, are you proposing that we consistently _don't_ escape `>` for all elements, unless it appears as part of `]]>` ?   I'm fine with that, but it's more of a change from the current behaviour.  I don't know of any good reason to escape `>`, but it does seem to be common practice.

> * In addition to the escape padding fix and \t\r\n improvement discussion, consider only conditionally escaping control characters based on 1.0.

Just to confirm, we're talking about the 7F-9F control characters here?

brettkail-wk: > To be clear, are you proposing that we consistently don't escape `>` for all elements, unless it appears as part of `]]> `?

That was my suggestion since it would remove all need to track `style`, but if you prefer to keep tracking `style` just for that divergence, that's fine with me.

> Just to confirm, we're talking about the 7F-9F control characters here?

Sorry, ignore that final bullet, I misremembered: XML 1.1 added support for 0x01-0x1f but only as character references, not literals.

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: I don't think we can distinguish between `<?pi ?>` and `<?pi?>`, but we can with `<?pi foo="bar" ?>` and `<?pi foo="bar"?>`.

I've made it collapse `<?pi ?>` to `<?pi?>` and preserve space otherwise.

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Done, but I've retained the escape mode tracking.  In the past we had a request to create a single-file viewer (JS embedded), which could also use this escape-mode tracking for `script` tags, if implemented, and this behaviour also minimises change from what we had before (although I can't see how it would make a difference).

brettkail-wk (dismissed):

sagesmith-wf (dismissed):

matthewbelisle-wf:
Hey @paulwarren-wk this seems fairly risky from a security perspective. Is there a simple command I can run to test various inputs and outputs? I want to poke at it but I've got a high severity vuln that is taking all my time for the next day or two.

matthewbelisle-wf:
Security +1 input text is generated server side, users have no control to abuse this with malformed tags or anything similar.

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,183 @@
...
-        fout.write(etree.tostring(xmlDocument, method="xml", encoding="utf-8", xml_declaration=True))
+    ESCAPES = {
+        ']]>': ']]&gt;',
+        '>': '&gt;',
+        '<': '&lt;',
+        '"': '&quot;',
+        '&': '&amp;'
+        }
+
+    MUST_ESCAPE_CHARS = r'<&\u0001-\u0008\u000B\u000C\u000E\u001F\u007F-\u009F'
```
matthewbelisle-wf, line 42: Not a blocker. I'm curious about these control chars like delete `\u007F` @paulwarren-wk . Is that included because it can actually delete the previous character in certain situations? I've never seen that happen but I wouldn't be surprised I guess, and from what I could tell the xml spec like https://www.w3.org/TR/xml/#NT-AttValue didn't care about these ranges.

iXBRLViewerPlugin/xhtmlserialize.py
```diff
@@ -13,29 +13,183 @@
...
+    def write_escape_text(self, s, escape_mode):
+        if s is None:
+            return
+        if escape_mode == EscapeMode.STYLE:
+            # HTML does not understand XML escapes inside a style element.
+            # Don't escape ">".  Escaping isn't required by XML, but is common
+            # practice, so we do it elsewhere.
+            # "&" and "<" may only appear in a string or comment in CSS, so escape
+            # as if they're in a string.
+            self.write(self.STYLE_ESCAPE_RE.sub(lambda m: self.escape_str(m[0]),s))
```
matthewbelisle-wf, line 115: Not a security risk, but it looks like CSS maybe has different escape syntax @paulwarren-wk ?

https://www.w3.org/International/questions/qa-escapes

I see there were some changes in the last 10 years, but latest chrome is giving me this.

```
>>> console.log(CSS.escape('<'));
\<
>>> console.log(CSS.escape('&'));
\&
>>> console.log(CSS.escape('\u0001'));
\1
```

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: 7F-9F are included because XML 1.1 [requires that they be escaped](https://www.w3.org/TR/2006/REC-xml11-20060816/#sec-xml11):

> In order to improve the robustness of character encoding detection, the additional control characters #x7F through #x9F, which were freely allowed in XML 1.0 documents, now must also appear only as character references.

It's not clear if XML 1.1 is permitted by iXBRL, but we agreed to try to support it in the serializer.

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: Ah - the comment there is out of date.  I've just committed an update.

The background here is that we're trying to serialize in a way that will yield a document that is both XHTML valid, and will render equivalently as XHTML and HTML, so we're trying to simultaneously comply with the different escaping rules applied by HTML and XHTML in `<style>` elements.

The issue that triggered this was that we were escaping `>` to `&gt;`, which is fine for XHTML, but not for HTML.

Escaping to `\<` and `\&` would still be XHTML invalid.  In earlier versions of this PR, we were escaping `<` and `&` using the `\NNNNNN` unicode syntax, on the basis that the only place those characters can currently legally appear are in strings, but we ultimately agreed to not do that, on the basis that a) there are proposals to add `&` to CSS syntax, and this escaping would break that and b) if HTML/XHTML compatibility was a goal of the original document author and these characters appeared in a string, they'd already be escaped.

iXBRLViewerPlugin/xhtmlserialize.py
paulwarren-wk: This was done in c3c45ad44a366a7150cda6dc0034a51b9915eb2a

iXBRLViewerPlugin/xhtmlserialize.py
matthewbelisle-wf: Okay, thanks for the info @paulwarren-wk .

iXBRLViewerPlugin/xhtmlserialize.py
matthewbelisle-wf: Okay, thanks for the info @paulwarren-wk . That new comment makes sense.

brettkail-wk (approved):

sagesmith-wf (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
QA +10 Unit and Skynet tests pass successfully. Builds successfully. Tested locally with a doc that has `>` and it was not escaped.

@Workiva/release-management-p

lukepint-wk:
QA +10

Testing Notes:
- After/Before screenshot of FERC form templates CSS:
Screen Shot 2023-03-02 at 9 48 02 AM
- After/Before screenshot of FERC form with fixed list styling:
Screen Shot 2023-03-02 at 10 02 51 AM

lukepint-wk (approved):

sagesmith-wf:
@matthewbelisle-wf, can I get you to refresh?

rmconsole-wf (approved):
+1 from RM

PR #389, opened 2023-02-25
Improve load performance

paulwarren-wk:
#### Reason for change

Reduce viewer loading time.

#### Description of change

This PR avoids two operations done during the pre-process iXBRL phase that trigger a reflow of the document.  On the attached test document on my laptop this reduces the pre-process phase from 18 seconds to 1 second.

* The `is(":hidden")` check, which is used to display the "concealed fact" warning in the inspector, is now done on demand.
* The check for zero-size nodes (because they contain absolutely positioned nodes) is now only done on elements that actually contain absolutely positioned nodes.  Many documents won't contain absolutely positioned content at all.  

#### Steps to Test

Try opening the attached IXDS in the viewer, and note the time taken to load before and after applying this PR:

[0000078003-20-000014-xbrl (1).zip](https://github.com/Workiva/ixbrl-viewer/files/10831330/0000078003-20-000014-xbrl.1.zip)

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -124,7 +124,7 @@ Viewer.prototype._wrapNode = function(n) {
     var wrapper = "<span>";
     const nn = n.getElementsByTagName("*");
     for (var i = 0; i < nn.length; i++) {
-        if($(nn[i]).css("display") === "block") {
+        if(nn[i].style.display === "block") {
             wrapper = '<div>';
             break;
         }
@@ -204,19 +204,22 @@ Viewer.prototype._findOrCreateWrapperNode = function(domNode) {
         nodes = this._wrapNode(domNode);
         // Create a node set of current node and all absolutely positioned
         // descendants.
-        nodes = nodes.find("*").addBack().filter(function () {
-            return (this == nodes[0] || $(this).css("position") == "absolute");
+        nodes = nodes.find("*").addBack().filter(function (n, e) {
+            // node list will include ix:* elements, with no style property.
+            // Can't trivially use instanceof HTMLElement in an iframe.
+            return (this == nodes[0] || ('style' in this && this.style.position === "absolute"));
         });
     }
     nodes.each(function (i) {
-        if (this.getBoundingClientRect().height == 0 && $(this).css('display') !== 'inline') {
-            $(this).addClass("ixbrl-no-highlight"); 
+        // getBoundingClientRect blocks on layout, so only do it if we've actually got absolute nodes
+        if (nodes.length > 1 && this.style.display !== 'inline' && this.getBoundingClientRect().height == 0) {
+            this.classList.add("ixbrl-no-highlight");
         }
         if (i == 0) {
-            $(this).addClass("ixbrl-element")
+            this.classList.add("ixbrl-element");
         }
         else {
-            $(this).addClass("ixbrl-sub-element"); 
+            this.classList.add("ixbrl-sub-element");
         }
     });
     return nodes;
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -124,7 +124,7 @@ Viewer.prototype._wrapNode = function(n) {
     var wrapper = "<span>";
     const nn = n.getElementsByTagName("*");
     for (var i = 0; i < nn.length; i++) {
-        if(nn[i].style.display === "block") {
+        if ('style' in nn[i] && nn[i].style.display === "block") {
             wrapper = '<div>';
             break;
         }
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -203,19 +204,22 @@ Viewer.prototype._findOrCreateWrapperNode = function(domNode) {
         nodes = this._wrapNode(domNode);
         // Create a node set of current node and all absolutely positioned
         // descendants.
-        nodes = nodes.find("*").addBack().filter(function () {
-            return (this == nodes[0] || $(this).css("position") == "absolute");
+        nodes = nodes.find("*").addBack().filter(function (n, e) {
+            // node list will include ix:* elements, with no style property.
+            // Can't trivially use instanceof HTMLElement in an iframe.
```
brettkail-wk, line 209: I'm not very familiar with jQuery usage here, so I'm not following the comment.  What does "with no style property" mean?  Why would you want to use `instanceof`, and what doesn't work?

paulwarren-wk: > What does "with no style property" mean

`this.style.foo` will fail, because `this.style` is undefined.

> Why would you want to use instanceof, and what doesn't work?

I wanted to use `this instanceof HTMLElement` as the guard, but that is never true on elements inside an iframe – you need to use `iframe.contentWindow.HTMLelement`, and we don't have convenient access to the relevant iframe object here.

brettkail-wk: Ah, so is the comment is justifying `'style' in this`?  That makes sense.  If you like, you could consider something like:
> The node list can contain non-HTMLElement ix:* nodes, so be defensive when checking for style.  In theory, we could use "instanceof HTMLElement" instead of "in", but that doesn't trivially work from an iframe.

lukepint-wk: > this.style.foo will fail, because this.style is undefined.

You could use [optional chaining](https://caniuse.com/mdn-javascript_operators_optional_chaining) instead `this.style?.foo` as long as that's enough browser support.

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -123,7 +124,7 @@ Viewer.prototype._wrapNode = function(n) {
     var wrapper = "<span>";
     const nn = n.getElementsByTagName("*");
     for (var i = 0; i < nn.length; i++) {
-        if($(nn[i]).css("display") === "block") {
+        if ('style' in nn[i] && nn[i].style.display === "block") {
```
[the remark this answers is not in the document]
paulwarren-wk: You're right - this should use `getComputedStyle`.  Now fixed.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: Yes, although it turns out that `getComputedStyle()` is less fussy - it only requires an `Element` as input, not an `HTMLElement`, so it doesn't object to the iXBRL nodes.

derekgengenbacher-wf (dismissed):

austinmatherne-wk (dismissed):

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -203,19 +204,20 @@ Viewer.prototype._findOrCreateWrapperNode = function(domNode) {
...
-            return (this == nodes[0] || $(this).css("position") == "absolute");
+        nodes = nodes.find("*").addBack().filter(function (n, e) {
+            return (this == nodes[0] || (getComputedStyle(this).getPropertyValue('position') === "absolute"));
         });
     }
     nodes.each(function (i) {
-        if (this.getBoundingClientRect().height == 0 && $(this).css('display') !== 'inline') {
-            $(this).addClass("ixbrl-no-highlight"); 
+        // getBoundingClientRect blocks on layout, so only do it if we've actually got absolute nodes
+        if (nodes.length > 1 && this.style.display !== 'inline' && this.getBoundingClientRect().height == 0) {
```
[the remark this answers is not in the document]
paulwarren-wk: Yes - you're right.  Fixed in [b467e51](https://github.com/Workiva/ixbrl-viewer/commit/b467e51c0e0e1925ed8811cd25d32e461acf816d)

derekgengenbacher-wf (approved):

brettkail-wk (approved):

austinmatherne-wk (approved):

derekgengenbacher-wf:
QA +10 Unit and Skynet tests pass successfully. Builds successfully. Tested locally, this dramatically reduces loading time.

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #392, opened 2023-03-07
Include version number in viewer

### LINKED ISSUE #368: [REQUEST] Add a version export button

The ixbrl-vewer bug reporting template asks for a version of the ixbrl-viewer to be included for easier debugging. Currently, to get the version of the ixbrl-viewer being used, a user would have to inspect the html document to find the script import which requires some technical abilities. Another issue with this is if the user is building their ixbrl-viewer using a locally built viewer then there is no discernible version included (`ixbrlviewer.js`). We should add a version export button that allows users to easily export the version of the ixbrl-viewer script that is being used and/or the githash used to build the viewer similar to how arelle does

paulwarren-wk:
#### Reason for change

Fixes #368 

#### Description of change

The viewer now shows the version number in the bottom right of the inspector.  The version number is obtained with `git describe --tags --dirty`, with `-dev` appended if it's a dev build.

There's nothing to stop the version number overlapping with the Workiva logo, but production builds should have a fairly short version number (e.g. `1.1.52`)

image

#### Steps to Test

Build viewer, open any document, confirm that version identifier is shown in bottom right corner.

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
   },
   "scripts": {
     "test": "jest",
-    "dev": "webpack --config iXBRLViewerPlugin/viewer/webpack.dev.js",
-    "prod": "webpack --config iXBRLViewerPlugin/viewer/webpack.prod.js",
+    "dev": "VIEWER_VERSION=`git describe --tags --dirty`-dev webpack --config iXBRLViewerPlugin/viewer/webpack.dev.js",
+    "prod": "VIEWER_VERSION=`git describe --tags --dirty` webpack --config iXBRLViewerPlugin/viewer/webpack.prod.js",
     "stylelint": "stylelint iXBRLViewerPlugin/viewer/src/less/*.less",
     "i18n": "i18next -c iXBRLViewerPlugin/viewer/i18next-parser.config.js"
   },
--- a/iXBRLViewerPlugin/viewer/src/html/inspector.html
+++ b/iXBRLViewerPlugin/viewer/src/html/inspector.html
@@ -179,6 +179,7 @@ limitations under the License.
     <div class="inspector-foot powered-by">
         <span data-i18n="inspector.poweredBy">Powered by</span>
         <a href="https://www.workiva.co.uk/solutions/xbrl-and-ixbrl" target="_blank"><img alt="Workiva logo (opens in new window)" src="../img/workiva.svg" /></a>
+        <div class="version"></div>
     </div>
     <div class="failed-to-load-mask"></div>
   </div>
--- a/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
@@ -99,6 +99,7 @@ iXBRLViewer.prototype._loadInspectorHTML = function () {
     $('<link id="ixv-favicon" type="image/x-icon" rel="shortcut icon" />')
         .attr('href', require('../img/favicon.ico'))
         .appendTo('head');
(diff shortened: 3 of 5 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

package.json
```diff
@@ -8,8 +8,8 @@
   },
   "scripts": {
     "test": "jest",
-    "dev": "webpack --config iXBRLViewerPlugin/viewer/webpack.dev.js",
-    "prod": "webpack --config iXBRLViewerPlugin/viewer/webpack.prod.js",
+    "dev": "VIEWER_VERSION=`git describe --tags --dirty`-dev webpack --config iXBRLViewerPlugin/viewer/webpack.dev.js",
```
sagesmith-wf, line 11: Does the plugin documentation need to be updated to make sure everyone includes this as part of their build? Or maybe it will just show an empty string?

austinmatherne-wk (approved):

package.json
paulwarren-wk: If the environment variable is not set, it results in an empty string, so there's nothing visible in the viewer.

sagesmith-wf: @paulwarren-wk FYI this only works correctly if the user has fetched all of the tags. So for me when I QA this prior to fetching everything:

Screenshot 2023-03-07 at 4 40 03 PM

v.s. after fetching everything

Screenshot 2023-03-07 at 4 44 36 PM

---

Its probably still an improvement but you might add `git fetch --tags` or something to your build command. I am also kind of iffy on if this will work well from forks who don't know about upstream, maybe thats an acceptable trade off though. I am fine with it as is, but lemme know if you don't want to change and I'll merge

paulwarren-wk: Interesting.  I wasn't aware that it was possible to do a `git fetch` without fetching tags, but it looks like there's a config option that makes not fetching tags the default.  Is this something you've selected?

I rather not add a `git fetch --tags` to the `dev` target, because it makes the build network-dependent, and even a no-op fetch can take a few seconds.  I'm also thinking that if someone has explicitly chosen to not fetch tags, then they may not want the build process to do so.

In your example, `1.1.38` may not be the most recent version number, but the `-70` does at least warn us that there have been quite a few changes since `1.1.38`.

> iffy on if this will work well from forks who don't know about upstream

I think it's only a problem if they start using their own tags that are confusable with our version numbers.  As long as they don't do that, then I think we'll just get an old version number and a large commit count, or whatever tags they choose to use on their fork.

sagesmith-wf: I'm doing `git fetch <YOUR_FORK` instead of `git fetch --tags` which I would say is fairly common for projects that have a lot of contributors

paulwarren-wk: > If the environment variable is not set, it results in an empty string, so there's nothing visible in the viewer.

Actually, there is an issue here which is that plugins will have their own webpack config, so unless they're updated to define `__VERSION__`, we'll get a `ReferenceError`.  I've just made a change to wrap the reference to `__VERSION__` in a `try`/`catch` to avoid this.

To get a version number showing in a plugin, the relevant webpack and build process will need to be updated to provide an appropriate definition of `__VERSION__`, so I think plugin authors are free to do that how they choose, although it would be useful to know what version of the viewer the plugin is built on - it should be possible to script `npm view ixbrl-viewer list` or similar when producing `__VERSION__` for a plugin.

sagesmith-wf (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

@Workiva/release-management-p

PR #394, opened 2023-03-16
Stub viewer mode

paulwarren-wk:
#### Reason for change

To reduce the delay between opening a report and the loading mask and progress information being displayed, which can be a significant amount of time on large documents.

This change also reduces the modifications made to the source document, as no `<script>` tags are inserted.  The only  changes made to the source document are the addition of `id` attributes on any `ix` elements that don't already have them.

#### Description of change

The change adds two new command line options:

* `--use-stub-viewer` - if specified, stub viewer mode is enabled.
* `--viewer-suffix` - a suffix that will be inserted into generated filenames.

When using stub viewer mode, the path given to `--save-viewer` should be a directory.  The viewer will output two files:

* `ixbrlviewer.html` - a small HTML file containing the `<script>` reference and JSON data.
* A file with the same name as the input file, with the viewer suffix (if specified) inserted before the file extension.

The viewer  is opened by opening the `ixbrlviewer.html` file.  Due to security restrictions, this needs to be open on a web server (this is the same restriction that applies to Inline XBRL document sets).

#### Steps to Test

Generate a viewer for any file using the command line and adding the `--use-stub-viewer` option and specifying a directory to `--save-viewer`.  Optionally specify a suffix, such as `--viewer-suffix=-viewer`.

This code has been in use on the filings.xbrl.org fork of the viewer for some time, and can be seen on any filing there (e.g. https://filings.xbrl.org/969500N6AVPA51648T62/2022-12-31/ESEF/FR/1/rexel-2022-12-31-fr/reports/ixbrlviewer.html)

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/__init__.py
+++ b/iXBRLViewerPlugin/__init__.py
@@ -42,8 +42,6 @@ import argparse
 import traceback
 import sys
 
-import traceback
-import sys
 
 def iXBRLViewerCommandLineOptionExtender(parser, *args, **kwargs):
     parser.add_option("--save-viewer",
@@ -62,6 +60,10 @@ def iXBRLViewerCommandLineOptionExtender(parser, *args, **kwargs):
     # Force logging to use a buffer so that messages are retained and can be
     # retrieved for inclusion with the viewer.
     parser.add_option("--logToBuffer", action="store_true", dest="logToBuffer", default=True, help=argparse.SUPPRESS)
+    parser.add_option("--use-stub-viewer",
+                      action="store_true",
+                      dest="useStubViewer",
+                      help="Use stub viewer for faster loading of inspector")
 
 def iXBRLViewerCommandLineXbrlRun(cntlr, options, *args, **kwargs):
     # extend XBRL-loaded run processing for this option
@@ -76,7 +78,7 @@ def iXBRLViewerCommandLineXbrlRun(cntlr, options, *args, **kwargs):
         out = getattr(options, 'saveViewerFile') or kwargs.get("responseZipStream")
         if out:
             viewerBuilder = IXBRLViewerBuilder(modelXbrl)
-            iv = viewerBuilder.createViewer(scriptUrl=options.viewerURL, showValidations = options.validationMessages)
+            iv = viewerBuilder.createViewer(scriptUrl=options.viewerURL, showValidations = options.validationMessages, useStubViewer=options.useStubViewer)
             if iv is not None:
                 iv.save(out, outBasenameSuffix=VIEWER_BASENAME_SUFFIX, outzipFilePrefix=VIEWER_BASENAME_SUFFIX)
     except IXBRLViewerBuilderError as ex:
--- a/iXBRLViewerPlugin/__init__.py
+++ b/iXBRLViewerPlugin/__init__.py
@@ -63,7 +63,7 @@ def iXBRLViewerCommandLineOptionExtender(parser, *args, **kwargs):
     parser.add_option("--use-stub-viewer",
                       action="store_true",
                       dest="useStubViewer",
-                      help="Use stub viewer for faster loading of inspector")
+                      help="Use stub viewer for faster loading of inspector (requires web server)")
     parser.add_option("--viewer-suffix",
                       action="store",
                       default="",
--- a/iXBRLViewerPlugin/__init__.py
+++ b/iXBRLViewerPlugin/__init__.py
@@ -64,6 +64,11 @@ def iXBRLViewerCommandLineOptionExtender(parser, *args, **kwargs):
                       action="store_true",
                       dest="useStubViewer",
                       help="Use stub viewer for faster loading of inspector")
+    parser.add_option("--viewer-suffix",
+                      action="store",
+                      default="",
+                      dest="viewerBasenameSuffix",
+                      help="Suffix for basename of viewer files")
 
 def iXBRLViewerCommandLineXbrlRun(cntlr, options, *args, **kwargs):
     # extend XBRL-loaded run processing for this option
@@ -80,7 +85,7 @@ def iXBRLViewerCommandLineXbrlRun(cntlr, options, *args, **kwargs):
             viewerBuilder = IXBRLViewerBuilder(modelXbrl)
             iv = viewerBuilder.createViewer(scriptUrl=options.viewerURL, showValidations = options.validationMessages, useStubViewer=options.useStubViewer)
             if iv is not None:
-                iv.save(out, outBasenameSuffix=VIEWER_BASENAME_SUFFIX, outzipFilePrefix=VIEWER_BASENAME_SUFFIX)
+                iv.save(out, outzipFilePrefix=VIEWER_BASENAME_SUFFIX)
     except IXBRLViewerBuilderError as ex:
         print(ex.message)
     except Exception as ex:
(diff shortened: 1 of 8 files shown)


aviary2-wf:
## Security Insights
<details>
<li>Watched keyword <code>etree</code> in <code>iXBRLViewerPlugin/iXBRLViewer.py</code> line(s) ['27', '339'] added</li>
</details>

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -165,7 +165,14 @@ iXBRLViewer.prototype.load = function () {
     setTimeout(function () {
 
         iv._loadInspectorHTML();
-        var iframes = $(iv._reparentDocument());
```
brettkail-wk, line 168: Out of curiosity, does the hardcoded `.html` force non-XHTML rendering quirks, or does this `_reparentDocument` already cause those to not work (require the input document to be written to the HTML/XHTML subset)?

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -99,7 +99,7 @@ function localName(e) {
 }
 
 Viewer.prototype._addDocumentSetTabs = function() {
-    if (this._report.isDocumentSet()) {
+    if (this._report.isDocumentSet() && this._report.documentSetFiles().length > 1) {
```
brettkail-wk, line 102: Why is this change required?  It looks like docSetFiles is the same for stub vs default.

iXBRLViewerPlugin/__init__.py
```diff
@@ -62,6 +60,15 @@ def iXBRLViewerCommandLineOptionExtender(parser, *args, **kwargs):
     # Force logging to use a buffer so that messages are retained and can be
     # retrieved for inclusion with the viewer.
     parser.add_option("--logToBuffer", action="store_true", dest="logToBuffer", default=True, help=argparse.SUPPRESS)
+    parser.add_option("--use-stub-viewer",
+                      action="store_true",
+                      dest="useStubViewer",
+                      help="Use stub viewer for faster loading of inspector")
```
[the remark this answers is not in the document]
paulwarren-wk: > Presumably the reason it's not on by default is due to security. 

Actually, it's just because you can no longer just open the resulting file and have it work.  Adding the note above is a good idea - now done.

> So to be sure, this doesn't reduce the delay for a user to interact with the document, it just causes the "loading" notification to appear more quickly, right?

Pretty much.  I think there is a small win from avoiding the "reparent" step, but it's primarily to get the loading mask up before rather than after the initial document load happens, which can be many seconds on a large document.

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
paulwarren-wk: The line above is from `master`, not the PR branch.   In stub viewer mode, there is no reparent step.  All documents are loaded via the `src` attribute, and I think rendering mode is determined by the content-type set by the web server, which typically follows the file extension.  So I think that even if the stub viewer is `.html`, the main file gets loaded as XHTML if it has a `.xhtml` extension.

In non-stub mode, I think the reparent step above results in the iframe inheriting rendering mode of the parent.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: In stub mode, docSetFiles is an array containing the iXBRL files, even if there's only one.  In normal mode, docSetFiles is undefined.

I've now moved the length check inside `isDocumentSet()`, and refactored the other code that used this method (30ae2ec) so `isDocumentSet` is back to its original meaning of "is a multi-document document set".

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -165,52 +165,63 @@ iXBRLViewer.prototype.load = function () {
...
-            var ds = report.documentSetFiles();
-            for (var i = 1; i < ds.length; i++) {
-                var iframe = $("<iframe />").attr("src", ds[i]).appendTo("#ixv #iframe-container");
-                iframes = iframes.add(iframe);
-            }
+        const report = new iXBRLReport(JSON.parse(taxonomyData));
+        const ds = report.documentSetFiles();
+        var hasExternalIframe = false;
+        for (var i = stubViewer ? 0 : 1; i < ds.length; i++) {
+            const iframe = $("<iframe />").attr("src", ds[i]).appendTo("#ixv #iframe-container");
```
brettkail-wk, line 187: Pre-existing, I think the `src` needs to somehow URL encode path segments.  Otherwise `Q3: Workiva 10-Q.htm` will be mishandled (`Q3` interpreted as a scheme and normalized to `q3`), or a filename like `javascript:alert(1).html` would become active content.

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
brettkail-wk: (Approved if you'd rather this be fixed separately.)

paulwarren-wk: Done in [f440403](https://github.com/Workiva/ixbrl-viewer/commit/f440403b388e326c78eea5dddfec3559478bf4a9)

I think it makes sense to do this in viewer generation, rather than the client, and treat this as a list of valid URLs, as this gives us a bit more flexibility in the future if we wanted to be able to put absolute URLs here.

soufianeelaissaoui-wk:
Looks good to me.
Aviary complained about etree, but seems like etree from lxml is safe.
Security +1

iXBRLViewerPlugin/__init__.py
```diff
@@ -62,6 +60,15 @@ def iXBRLViewerCommandLineOptionExtender(parser, *args, **kwargs):
     # Force logging to use a buffer so that messages are retained and can be
     # retrieved for inclusion with the viewer.
     parser.add_option("--logToBuffer", action="store_true", dest="logToBuffer", default=True, help=argparse.SUPPRESS)
+    parser.add_option("--use-stub-viewer",
+                      action="store_true",
+                      dest="useStubViewer",
+                      help="Use stub viewer for faster loading of inspector (requires web server)")
+    parser.add_option("--viewer-suffix",
```
derekgengenbacher-wf, line 67: Should we add these options to the build_viewer.py file?

brettkail-wk (dismissed):

iXBRLViewerPlugin/__init__.py
paulwarren-wk: Done in [c8a59dd](https://github.com/Workiva/ixbrl-viewer/pull/394/commits/c8a59dd60ca61c8619898a2c7a422c377c83f6e7)

derekgengenbacher-wf (approved):

paulyun-wk (approved):
Security +1
for latest commit

austinmatherne-wk (approved):

rmconsole-wf (approved):
+1 from RM

PR #401, opened 2023-03-23
Improve load time

paulwarren-wk:
#### Reason for change

The load process can be slow on large documents, preventing the user from interacting with the viewer.

#### Description of change

This PR moves two slow operations into a asynchronous "post load" phase after the loading mask is removed, and the viewer becomes interactive.  These are:

* Checking for zero-size elements.  This check avoids putting an outline around empty elements.  This is slow, because it relies on the layout-triggering `getClientBoundingRect()`.  In the event that a user selects an affected fact before this operation is completed, the results is purely cosmetic, and will be fixed as soon as the worker gets to that element.
* Building the search index.  Search is now disabled until this operation is complete. 

Working with [this UBS filing](https://www.sec.gov/Archives/edgar/data/1114446/000161052022000028/0001610520-22-000028-index.html), I see the "preprocessing" step reduced from about 12s to 6s, and the build search index phase of about 5s is also deferred.

#### Steps to Test

Open a big filing, such as the UBS one above, and compare time from opening to loading mask disappearing.  Check behaviour of search function immediately after load.

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/html/inspector.html
+++ b/iXBRLViewerPlugin/viewer/src/html/inspector.html
@@ -57,7 +57,7 @@ limitations under the License.
 
       <div class="search-controls search-mode-only">
         <div class="search-input">
-          <input id="ixbrl-search" type="text" aria-label="Search iXBRL report"/>
+          <input id="ixbrl-search" type="text" aria-label="Search iXBRL report" disabled="disabled" />
           <div class="filter-toggle"></div>
         </div>
         <div class="search-filters">
@@ -162,7 +162,8 @@ limitations under the License.
         </div>
         <div class="search-overlay">
           <div class="search-overlay-icon"></div>
-          <div class="search-overlay-text"><p class="title" data-i18n="inspector.factSearch">Fact Search</p><p class="text">Please enter some search terms</p></div>
+          <div class="search-overlay-text search-ready-only"><p class="title" data-i18n="inspector.factSearch">Fact Search</p><p class="text">Please enter some search terms</p></div>
+          <div class="search-overlay-text search-not-ready-only"><p class="title" data-i18n="inspector.factSearch">Fact Search</p><p class="text">Please wait - building index...</p></div>
         </div>
       </div>
 
--- a/iXBRLViewerPlugin/viewer/src/js/search.js
+++ b/iXBRLViewerPlugin/viewer/src/js/search.js
@@ -18,10 +18,10 @@ import $ from 'jquery'
 export class ReportSearch {
     constructor(report) {
         this._report = report;
-        this.buildSearchIndex();
+        this.ready = false;
     }
 
-    buildSearchIndex() {
+    * buildSearchIndex(doneCallback) {
         var docs = [];
         var dims = {};
         var facts = this._report.facts();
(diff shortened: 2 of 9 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/html/inspector.html
```diff
@@ -57,7 +57,7 @@ <h1 class="footnote-mode-only" data-i18n="inspector.footnoteProperties">Footnote
 
       <div class="search-controls search-mode-only">
         <div class="search-input">
-          <input id="ixbrl-search" type="text" aria-label="Search iXBRL report"/>
+          <input id="ixbrl-search" type="text" aria-label="Search iXBRL report" disabled="disabled" />
```
lukepint-wk, line 60: `disabled="disabled"` should just be `disabled` I think since it's a boolean attribute.

paulwarren-wk: Yes, that's correct.  I know that we have had issues with viewer files not being XHTML-compatible in the past, so I've been aiming for XHTML-compatible throughout.  Interestingly, browsers don't seem to care about this issue here (`disabled` vs `disabled="disabled"`) but they do complain about using `&nbsp;` if you open the viewer in XHTML mode.

derekgengenbacher-wf (dismissed):

austinmatherne-wk (approved):

derekgengenbacher-wf:
QA +10 Unit and Skynet tests pass successfully. Builds successfully. This dramatically reduces the load time of the provided doc.

@Workiva/release-management-p

iXBRLViewerPlugin/viewer/src/html/inspector.html
lukepint-wk: That makes sense, thanks for the explanation!

sagesmith-wf (dismissed):

brettkail-wk (approved):

austinmatherne-wk (approved):

derekgengenbacher-wf (dismissed):

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/js/search.js
```diff
@@ -65,6 +65,16 @@ export class ReportSearch {
...
         }
         const builder = new lunr.Builder();
+        builder.pipeline.add(
+          lunr.trimmer,
+          lunr.stopWordFilter,
+          lunr.stemmer
+        )
+
+        builder.searchPipeline.add(
+          lunr.stemmer
```
austinmatherne-wk, line 75: Any particular reason not to also use the trimmer and stop word pipeline functions for search?

paulwarren-wk: I'm not sure.  This is just replicating the default behaviour that you get if you use the `lunr()` convenience constructor (see https://github.com/olivernn/lunr.js/blob/master/lunr.js#L43) which is what we were using before this PR.  I assume the idea is that people wouldn't use punctuation in their search queries.  

We had to move away from the convenience constructor because it uses a callback to populate the index, and we can't `yield` from within the callback.

austinmatherne-wk (dismissed):

derekgengenbacher-wf:
Did some additional testing and functionality still looks good, but the consumer testing continues to pick up some differences with the search panel. Ive pass these differences along to Paul offline.

rmconsole-wf (approved):
+1 from RM

PR #407, opened 2023-04-24
XT-2866 Add ability to filter positive/negative fact values

sagesmith-wf:
#### Reason for change
General improvements

#### Description of change
Adds a fact value positive/negative filter

Screenshot 2023-04-24 at 2 13 18 PM

#### Steps to Test
- Pull this branch
- Build the javascript
  - `npm install && npm run prod`
- Create a filing with some positive an negative facts associated
- From arelle run: `python arelleCmdLine.py --plugins=/PATH_TO/ixbrl-viewer/iXBRLViewerPlugin -f /PATH_TO/filing_documents/ixbrldoc.htm  --save-viewer ixbrl-report-viewer.html --viewer-url /PATH_TO/ixbrl-viewer/iXBRLViewerPlugin/viewer/dist/ixbrlviewer.js`
- `open ixbrldoc.htm`

### NOTE:
`--save-viewer` arg doesn't seem to be taken into account when generating a file

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/search.js
+++ b/iXBRLViewerPlugin/viewer/src/js/search.js
@@ -14,6 +14,7 @@
 
 import lunr from 'lunr'
 import $ from 'jquery'
+import Decimal from "decimal.js";
 
 export class ReportSearch {
     constructor(report) {
@@ -108,10 +109,11 @@ export class ReportSearch {
 
         rr.forEach((r,i) => {
                 var item = searchIndex._report.getItemById(r.ref);
-                if (
-                    (item.isHidden() ? s.showHiddenFacts : s.showVisibleFacts) &&
-                    (s.periodFilter == '*' || item.period().key() == s.periodFilter) &&
-                    (s.conceptTypeFilter == '*' || s.conceptTypeFilter == (item.isNumeric() ? 'numeric' : 'text'))) {
+                if ((item.isHidden() ? s.showHiddenFacts : s.showVisibleFacts) &&
+                    (s.periodFilter === '*' || item.period().key() === s.periodFilter) &&
+                    (s.conceptTypeFilter === '*' || s.conceptTypeFilter === (item.isNumeric() ? 'numeric' : 'text')) &&
+                    (s.factValueFilter === '*' || (s.factValueFilter === 'positive' && item.isPositive()) || (s.factValueFilter === 'negative' && item.isNegative())))
+                {
                     results.push({
                         "fact": item,
                         "score": r.score
--- a/iXBRLViewerPlugin/viewer/src/js/search.test.js
+++ b/iXBRLViewerPlugin/viewer/src/js/search.test.js
@@ -1,0 +1,142 @@
+// Copyright 2023 Workiva Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { ReportSearch } from "./search.js"
+import {iXBRLReport} from "./report";
+
+const testReportData = {
+    "concepts": {
+        "us-gaap:Cash": {
+            "labels": {
+                "ns0": {
+                    "en-us": "Cash"
+                }
+            }
+        }
+    },
+    "languages": {
+        "en-us": "En (US)"
+    },
+    "facts": {},
+    "prefixes": {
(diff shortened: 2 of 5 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/search.js
```diff
@@ -108,10 +109,11 @@ export class ReportSearch {
 
         rr.forEach((r,i) => {
                 var item = searchIndex._report.getItemById(r.ref);
-                if (
-                    (item.isHidden() ? s.showHiddenFacts : s.showVisibleFacts) &&
+                if ((item.isHidden() ? s.showHiddenFacts : s.showVisibleFacts) &&
                     (s.periodFilter == '*' || item.period().key() == s.periodFilter) &&
-                    (s.conceptTypeFilter == '*' || s.conceptTypeFilter == (item.isNumeric() ? 'numeric' : 'text'))) {
+                    (s.conceptTypeFilter == '*' || s.conceptTypeFilter == (item.isNumeric() ? 'numeric' : 'text')) &&
+                    (s.factValueFilter == '*') || (item.isNumeric() && !item.isNil() && s.factValueFilter == (new Decimal(item.value()).isPositive() ? 'positive': 'negative')))
```
brettkail-wk, line 115: The parentheses look wrong; I think it needs to be `(s.factValueFilter == '*' || ...)` or else the value filter won't intersect with the other filters.  I don't see any pre-existing tests for search, but it seems like that would be good to add at some point.

Is `new Decimal` correct, or should it be `Decimal` (as in `formatNumber`)?  Do we need an `undefined` test?  Perhaps we should add a helper method to fact to do this.  You might consider naming the method `isNegative()` since `isPositive` returning true for 0 seems somewhat confusing to me.

sagesmith-wf: > The parentheses look wrong; I think it needs to be (s.factValueFilter == '*' || ...) or else the value filter won't intersect with the other filters

Good catch, thanks

> I don't see any pre-existing tests for search, but it seems like that would be good to add at some point.

I can attempt this

> Is new Decimal correct, or should it be Decimal (as in formatNumber)?

I looked at `formatNumber` and it will return a bunch of things that aren't actually a decimal which is why I decided to just check the conditions for being a valid `Decimal`  being returned upfront and pass it in directly so that I am guaranteed to have access to `isPositive()`

>  Do we need an undefined test?

Maybe? I don't know what that explicitly refers to though

> Perhaps we should add a helper method to fact to do this.

Seems reasonable to me.

> You might consider naming the method isNegative()

Roger that.

sagesmith-wf: I didn't resolve the fact helper question mostly because I didn't know what it would look like ideally

iXBRLViewerPlugin/viewer/src/js/search.test.js
```diff
@@ -0,0 +1,143 @@
+// Copyright 2019 Workiva Inc.
```
austinmatherne-wk, line 1: nit: 2023

austinmatherne-wk (approved):

aaroncameron-wk (approved):

iXBRLViewerPlugin/viewer/src/js/search.js
paulwarren-wk: > You might consider naming the method isNegative() since isPositive returning true for 0 seems somewhat confusing to me.

I agree that the method should be named precisely, but what is the desired end user behaviour here?  If a user selects "positive" in the filter, do they want zeros included?  I'm guessing the use case here is for reviewing facts that have been tagged with the wrong sign, in which case you probably want to ignore zeros for both "positive" and "negative" filters.

FWIW, XML Schema uses "nonNegative" and "nonPositive" in datatype names order to unambiguously include zeros.

iXBRLViewerPlugin/viewer/src/js/search.test.js
```diff
@@ -0,0 +1,143 @@
...
+        "v": undefined,
+        "a": {
+            "c": "us-gaap:Cash",
+            "u": "iso4217:USD",
+            "p": "2018-01-01/2019-01-03",
+        }};
+    var report = testReport({'positive': positive, 'negative': negative, 'zero': zero, 'text': text, 'undefined': undef}, {'positive': {}, 'negative': {}, 'zero': {}, 'text': {}, 'undefined': {}})
+    var reportSearch = new ReportSearch(report);
+    var searchIndex = reportSearch.buildSearchIndex(() => {});
+    while (!reportSearch.ready) {
```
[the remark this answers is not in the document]
sagesmith-wf: Thank you!

aaroncameron-wk (approved):

allenbross-wf (approved):

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

lukepint-wk (approved):

austinmatherne-wk:
# Verified

* new filter locally 

QA +10
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #415, opened 2023-05-04
XT-2867: Filter facts by calculations

aaroncameron-wk:
#### Description of change
Filter facts based on their concept's status as:
- Calculation summation item
- Calculation contributing item
- Either of the above

Reworked the code that applies the filter into individual functions rather than a single if statement that will grow in complexity.
Added some helper functions to the search test code.
Screenshot 2023-05-04 at 2 36 01 PM
Screenshot 2023-05-04 at 2 36 11 PM

#### Steps to Test
- CI
- Load a filing of your choosing and ensure each calculation filter works as you would expect

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/html/inspector.html
+++ b/iXBRLViewerPlugin/viewer/src/html/inspector.html
@@ -77,9 +77,9 @@ limitations under the License.
           </div>
           <select id="search-filter-calculations">
             <option value="*" data-i18n="inspector.allOption">ALL</option>
-            <option value="contributing" data-i18n="inspector.contributing">Contributing Item</option>
-            <option value="summation" data-i18n="inspector.summation">Summation Item</option>
-            <option value="either" data-i18n="inspector.either">Either</option>
+            <option value="summation" data-i18n="inspector.summation">Summation</option>
+            <option value="contributor" data-i18n="inspector.contributor">Contributor</option>
+            <option value="summationOrContributor" data-i18n="inspector.summationOrContributor">Summation or Contributor</option>
           </select>
 
           <div class="control-label" data-i18n="inspector.factValue">
--- a/iXBRLViewerPlugin/viewer/src/html/inspector.html
+++ b/iXBRLViewerPlugin/viewer/src/html/inspector.html
@@ -72,6 +72,16 @@ limitations under the License.
               <option value="text" data-i18n="inspector.textOption">Text</option>
           </select>
 
+          <div class="control-label" data-i18n="inspector.calculations">
+            Calculations
+          </div>
+          <select id="search-filter-calculations">
+            <option value="*" data-i18n="inspector.allOption">ALL</option>
+            <option value="contributing" data-i18n="inspector.contributing">Contributing Item</option>
+            <option value="summation" data-i18n="inspector.summation">Summation Item</option>
+            <option value="either" data-i18n="inspector.either">Either</option>
+          </select>
+
           <div class="control-label" data-i18n="inspector.factValue">
               Fact Value
           </div>
--- a/iXBRLViewerPlugin/viewer/src/html/inspector.html
+++ b/iXBRLViewerPlugin/viewer/src/html/inspector.html
@@ -67,7 +67,7 @@ limitations under the License.
               Concept Type
           </div>
           <select id="search-filter-concept-type">
-              <option value="*" data-i18n="inspector.allOption">ALL</option>
+              <option value="*" data-i18n="inspector.noFilter">-</option>
               <option value="numeric" data-i18n="inspector.numericOption">Numeric</option>
               <option value="text" data-i18n="inspector.textOption">Text</option>
           </select>
@@ -76,7 +76,7 @@ limitations under the License.
             Calculations
           </div>
           <select id="search-filter-calculations">
-            <option value="*" data-i18n="inspector.allOption">ALL</option>
+            <option value="*" data-i18n="inspector.noFilter">-</option>
             <option value="summation" data-i18n="inspector.summation">Summation</option>
             <option value="contributor" data-i18n="inspector.contributor">Contributor</option>
             <option value="summationOrContributor" data-i18n="inspector.summationOrContributor">Summation or Contributor</option>
@@ -86,7 +86,7 @@ limitations under the License.
               Fact Value
           </div>
           <select id="search-filter-fact-value">
-              <option value="*" data-i18n="inspector.allOption">ALL</option>
+              <option value="*" data-i18n="inspector.noFilter">-</option>
               <option value="negative" data-i18n="inspector.negative">Negative</option>
               <option value="positive" data-i18n="inspector.positive">Positive</option>
           </select>
--- a/iXBRLViewerPlugin/viewer/src/js/report.js
+++ b/iXBRLViewerPlugin/viewer/src/js/report.js
@@ -72,29 +72,29 @@ iXBRLReport.prototype._initialize = function () {
 iXBRLReport.prototype.isCalculationContributor = function(c) {
     if (this._calculationContributors === undefined) {
         if (this.data.rels?.calc) {
-            this._calculationContributors = Object.values(this.data.rels.calc).flatMap(calculations => {
+            this._calculationContributors = new Set(Object.values(this.data.rels.calc).flatMap(calculations => {
                 return Object.values(calculations).flatMap(contributors => {
                     return contributors.map(c => c.t);
                 });
-            });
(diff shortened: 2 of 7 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/html/inspector.html
```diff
@@ -72,6 +72,16 @@ <h3 data-i18n="inspector.filter">Filter</h3>
               <option value="text" data-i18n="inspector.textOption">Text</option>
           </select>
 
+          <div class="control-label" data-i18n="inspector.calculations">
+            Calculations
+          </div>
+          <select id="search-filter-calculations">
+            <option value="*" data-i18n="inspector.allOption">ALL</option>
```
brettkail-wk, line 79: The distinction between "ALL" and "Either" seems somewhat confusing.  I wonder if empty string or "(No filter)" would be clearer than "ALL" (here and above), and I wonder if explicitly "Summation or Contributor" would be clearer than "Either".

Calculations are defined in terms of the summation, so it might be more natural to list "Summation" first. (In calculation 1.1, the summation is called total, but it seems fine to keep it as-is for now.)

I'm not sure "Item" part adds much, so you might consider dropping it.

iXBRLViewerPlugin/viewer/src/js/report.js
```diff
@@ -69,6 +69,22 @@ iXBRLReport.prototype._initialize = function () {
     }
 }
 
+iXBRLReport.prototype.isCalculationContributingConcept = function(c) {
+    if (!this.data.rels?.calc) return false;
+    return Boolean(Object.values(this.data.rels.calc).some(calculations => {
+        return Object.values(calculations).some(contributingItems => {
+            return contributingItems.some(i => i.t === c);
```
brettkail-wk, line 76: In thinking about startup time, this appears to be O(facts*contributors).  It seems like we could compute a set of contributors so that this method can just be `this.contributingItems.has(c)`, and similarly for summations.  We could change fact.isCalcSum/item to be methods rather than fields (or to be lazily computed fields), and then change this to be a lazily computed set of contributors to avoid affecting startup at all.

iXBRLViewerPlugin/viewer/src/js/search.js
```diff
@@ -107,12 +144,17 @@ export class ReportSearch {
...
+        ];
+
         rr.forEach((r,i) => {
                 var item = searchIndex._report.getItemById(r.ref);
-                if ((item.isHidden() ? s.showHiddenFacts : s.showVisibleFacts) &&
-                    (s.periodFilter === '*' || item.period().key() === s.periodFilter) &&
-                    (s.conceptTypeFilter === '*' || s.conceptTypeFilter === (item.isNumeric() ? 'numeric' : 'text')) &&
-                    (s.factValueFilter === '*' || (s.factValueFilter === 'positive' && item.isPositive()) || (s.factValueFilter === 'negative' && item.isNegative())))
+                if (filters.every(f => f(s, item)))
                 {
```
brettkail-wk, line 158: nit: pre-existing, suggest moving this `{` to the preceding line as it was before `factValueFilter`.

iXBRLViewerPlugin/viewer/src/js/report.js
```diff
@@ -69,20 +69,32 @@ iXBRLReport.prototype._initialize = function () {
...
-
-iXBRLReport.prototype.isCalculationSummationConcept = function(c) {
-    if (!this.data.rels?.calc) return false;
-    return Boolean(Object.values(this.data.rels.calc).some(calculations => {
-        return Object.keys(calculations).includes(c);
-    }));
+iXBRLReport.prototype.isCalculationContributor = function(c) {
+    if (this._calculationContributors === undefined) {
+        if (this.data.rels?.calc) {
+            this._calculationContributors = Object.values(this.data.rels.calc).flatMap(calculations => {
```
brettkail-wk, line 75: Suggest wrapping in  `new Set` and use `.has` below.  Otherwise, this is still O(facts*contributors), just deferred until the first use of the filter.

With that, do we need the pseudo-private fact field properties at all, or can the method just forward?  I don't know much about JavaScript, but I've read that engines like to optimize based on the shape (fields) of an object, so I don't know the effect on dynamically adding a property.

iXBRLViewerPlugin/viewer/src/html/inspector.html
aaroncameron-wk: Agreed and applied.

For "ALL", I felt that just `-` was better than an empty string, it gives something more clear to click on in the dropdown (at least the way it's rendered on Mac), but this certainly isn't my expertise so I'll concede if anyone has strong opinions.

iXBRLViewerPlugin/viewer/src/js/report.js
```diff
@@ -69,6 +69,34 @@ iXBRLReport.prototype._initialize = function () {
...
+iXBRLReport.prototype.isCalculationContributor = function(c) {
+    if (this._calculationContributors === undefined) {
+        if (this.data.rels?.calc) {
+            this._calculationContributors = new Set(Object.values(this.data.rels.calc).flatMap(calculations => {
+                return Object.values(calculations).flatMap(contributors => {
+                    return contributors.map(c => c.t);
+                });
+            }));
+        } else {
+            this._calculationContributors = [];
```
austinmatherne-wk, line 81: Both here and in `isCalculationSummation`. The no calculations case should be an empty set. Otherwise the `has(...)` calls will throw type errors.

austinmatherne-wk (approved):

iXBRLViewerPlugin/viewer/src/js/search.js
```diff
@@ -99,21 +99,62 @@ export class ReportSearch {
         doneCallback();
     }
 
+    visibilityFilter(s, item) {
+        return item.isHidden() ? s.showHiddenFacts : s.showVisibleFacts;
+    }
+
+    periodFilter(s, item) {
```
sagesmith-wf, line 106: This is fine, but since you refactored something like this might be more generic:

```python
class ValueMatcher:
    name: str
    value: str
    func: Callable

class Filter:
    def __init__(name, value_matchers):
        self.name = name
        self.value_matchers = {
            v.value: v for v in value_matchers
        }

    def html():
         return f'<div class="control-label" data-i18n="inspector.{self.name.lower()}">' +
         self.name +
          '</div>' +
          f'<select id="search-filter-{self.name}">' +
          '\n'.join([f'<option value="{v.value}" data-i18n="inspector.{v.value}">{v.name}</option>' for v in self.value_matchers]) +
          '</select>'
         
    def applies(filterValue, item):
        return filters[value].func(item)
...
filters = [Filter('Concept Type', [
     ValueMatcher('ALL', '*': lambda item: return True),
     ValueMatcher('NUMERIC', 'numeric': lambda item: item.isNumeric()),
     ValueMatcher('TEXT', 'text': lambda item: not item.isNumeric())
])]
```

sagesmith-wf (approved):

iXBRLViewerPlugin/viewer/src/js/search.js
aaroncameron-wk: This works well for dropdown filters, but not the visible/hidden checkboxes, and I anticipate some soon-to-be-implemented filters that might be more complex, like multi-select. I think once we have all those filters implemented we will be in a better place to revisit a larger refactor like this that takes over HTML generation.

sagesmith-wf: Yeah works for me

iXBRLViewerPlugin/viewer/src/html/inspector.html
```diff
@@ -72,6 +72,16 @@ <h3 data-i18n="inspector.filter">Filter</h3>
...
+            <option value="contributing" data-i18n="inspector.contributing">Contributing Item</option>
+            <option value="summation" data-i18n="inspector.summation">Summation Item</option>
+            <option value="either" data-i18n="inspector.either">Either</option>
```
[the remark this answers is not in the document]
brettkail-wk: It's not the same, but I think your sentiment is the same as [this comment](https://github.com/Workiva/ixbrl-viewer/pull/415#discussion_r1185514579) (fixed in followup).

austinmatherne-wk:
# Verified

* local testing of search filters

QA +10
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #420, opened 2023-05-10
XT-2861 XT-2862 Document Summary Pane

austinmatherne-wk:
#### Reason for change
Request to add document summary information to viewer.

Potential follow ups:
* hovering over namespace prefixes displays a tooltip containing the full namespace.
* file summary provides download links.

#### Description of change
Screenshot 2023-05-09 at 10 16 44 PM

#### Steps to Test
* Generate a viewer and confirm the details in the summary pane are correct.
* Verify that the summary details pane is usable within the inspector at a variety of widths.

**review**:
@Workiva/xt
@paulwarren-wk


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -403,6 +411,33 @@ def createViewer(self, scriptUrl=DEFAULT_VIEWER_PATH, useStubViewer=False, showV
...
             self.taxonomyData["filingDocuments"] = filingDocZipName
 
+        localDocs = defaultdict(set)
+        for path, doc in dts.urlDocs.items():
+            if isHttpUrl(path):
+                continue
+            if doc.type == Type.INLINEXBRL:
+                localDocs[doc.basename].add('inline')
+            elif doc.type == Type.INLINEXBRLDOCUMENTSET:
+                localDocs[doc.basename].add('docset')
```
brettkail-wk, line 421: I'm not familiar enough with Arelle.  What does `docset` vs `inline` mean?  It seems like they should just be the same for ixbrl-viewer?

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -403,6 +411,33 @@ def createViewer(self, scriptUrl=DEFAULT_VIEWER_PATH, useStubViewer=False, showV
...
+                    if child.qname == XbrlConst.qnLinkCalculationLink:
+                        localDocs[doc.basename].add('calcLinkbase')
+                    if child.qname == XbrlConst.qnLinkDefinitionLink:
+                        localDocs[doc.basename].add('defLinkbase')
+                    elif child.qname == XbrlConst.qnLinkLabelLink:
+                        localDocs[doc.basename].add('labelLinkbase')
+                    elif child.qname == XbrlConst.qnLinkPresentationLink:
+                        localDocs[doc.basename].add('presLinkbase')
+                    elif child.qname == XbrlConst.qnLinkReferenceLink:
+                        localDocs[doc.basename].add('refLinkbase')
```
brettkail-wk, line 435: I wonder if we should add a final `else .add('linkbase')` and corresponding "Other Linkbases" (possibly only enabled when non-zero)?

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -179,9 +180,16 @@ def addConcept(self, concept, dimensionType = None):
...
 
-            if concept.type.isTextBlock:
+            if concept.type is not None and concept.type.isTextBlock:
                 conceptData['t'] = True
 
+            if concept.isTypedDimension:
+                typedDomainElement = concept.typedDomainElement
+                if typedDomainElement is not None:
+                    typedDomainName = self.nsmap.qname(typedDomainElement.qname)
+                    conceptData['td'] = typedDomainName
```
brettkail-wk, line 190: Thoughts on data backwards compatibility (i.e., old JSON, new viewer.js)?  I wonder if we should add a top-level data version key, and the viewer.js can disable features for older generated.

iXBRLViewerPlugin/viewer/src/js/summary.js
```diff
@@ -0,0 +1,155 @@
...
+    _buildTagCounts() {
+        const tagCounts = new Map();
+        for (const fact of this._report.facts()) {
+            let counter = this._getTagCounter(tagCounts, fact.conceptName());
+            counter.addPrimaryItem(fact.conceptName());
+            for (const [dimension, member] of Object.entries(fact.dimensions())) {
+                counter = this._getTagCounter(tagCounts, dimension);
+                counter.addDimension(dimension);
+
+                if (member.indexOf(":") > -1) {
```
brettkail-wk, line 108: Doesn't this miscount typed dimension values that happen to contain `:`?  It seems like this block should somehow be in an `else` of `isTypedDimension`?

iXBRLViewerPlugin/viewer/src/js/summary.js
```diff
@@ -0,0 +1,155 @@
...
+                    const typedDomainElement = dimensionConcept.typedDomainElement();
+                    if (typedDomainElement) {
+                        counter = this._getTagCounter(tagCounts, typedDomainElement);
+                        counter.addMember(typedDomainElement);
+                    }
+                }
+            }
+        }
+        this._tagCounts = new Map(
+                [...tagCounts].map(([k, v]) => [k, v.getCounts()])
```
brettkail-wk, line 124: nit: indentation

iXBRLViewerPlugin/iXBRLViewer.py
austinmatherne-wk: You're right. The distinction from within the viewer is unimportant. I've combined them.

iXBRLViewerPlugin/iXBRLViewer.py
austinmatherne-wk: Done. I suppose it's worth pointing out these are all optional. The various file summary sections only display [if there are files under them](https://github.com/austinmatherne-wk/ixbrl-viewer/blob/a314f360b795cf6780d0b09511719692a1545076/iXBRLViewerPlugin/viewer/src/js/inspector.js#L446).

iXBRLViewerPlugin/iXBRLViewer.py
austinmatherne-wk: For now I'm intentionally making sure we handle the unset state of new fields from the js side. The process of field removal is rather more complicated, considering that we allow folks to create plugins on top of the viewer.

It seems reasonable to me to version the JSON data alongside the project (as the JSON is akin to our library API for those extending the viewer). From a semver perspective, field removal or type change would necessitate a major version change, new fields would lead to minor changes, and other modifications would be deemed patches. Thoughts?

iXBRLViewerPlugin/viewer/src/js/summary.js
austinmatherne-wk: Good catch. Updated.

iXBRLViewerPlugin/viewer/src/js/summary.test.js
```diff
@@ -0,0 +1,60 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { DocumentSummary } from "./summary.js";
+
+
+function testFact(conceptName, dimensions) {
```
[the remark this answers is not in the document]
austinmatherne-wk: I'll write up a ticket

iXBRLViewerPlugin/iXBRLViewer.py
brettkail-wk: > For now I'm intentionally making sure we handle the unset state of new fields from the js side.

Apologies, yes, I see that now.  I originally thought this was for `isTypedDimension() { return 'td' in this._c; }`, which would be wrong (absence of `td` means old JSON or non-typed-dimension), but that's not what the PR does.  (With that wrong assumption, I was trying to ask if we needed `if ('td' in fact._c) { typed } else if (dataVersion greater-than something) { explicit } else { unknown }`.)

> It seems reasonable to me to version the JSON data alongside the project (as the JSON is akin to our library API for those extending the viewer)

Agreed, we have [JavaScript Versioning](https://github.com/Workiva/ixbrl-viewer#javascript-versioning) section, and I think that's consistent with what you're saying.

sagesmith-wf (approved):

aaroncameron-wk (dismissed):

iXBRLViewerPlugin/viewer/src/less/summary.less
```diff
@@ -0,0 +1,67 @@
...
+
+      td,
+      th {
+        padding: 0.2rem;
+        text-align: left;
+      }
+
+      thead {
+        th {
+          word-break: break-word;
```
aaroncameron-wk, line 47: So this isn't a critical UI issue, but here are some options on getting the summary table to work a bit narrower:
Apparently `word-break: break-word;` [is deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), so the headers are breaking mid-word. The alternative would be:
```suggestion
          white-space: break-spaces;
```
But that means only "Primary Items" breaks a line, so the minimum table width is still larger than the starting inspector width. We could do one or both of these options:

1. Wrap the table in `<div style="overflow-x:auto">...</div>` so that when the table does exceed the inspector width, it can still be fully viewed.
2. Abbreviate some/all of the headers (NS, PI, Dim, Mem, etc.)

Screenshot 2023-05-11 at 1 09 37 PM
Screenshot 2023-05-11 at 1 09 57 PM
Screenshot 2023-05-11 at 1 11 03 PM
Screenshot 2023-05-11 at 1 13 42 PM

aaroncameron-wk (approved):

aaroncameron-wk:
QA +10
CI passes. Values match. Looks good!

rmconsole-wf (approved):
+1 from RM

PR #421, opened 2023-05-12
XT-2873: Units filter

aaroncameron-wk:
#### Reason for change
Filter facts by complex units.

#### Description of change
This involved refactoring the current representation of "units" into "measures", and "units" now refers to the complex unit.

#### Steps to Test
Unit tests added.
Open doc and confirm expected results for:
- No unit selected
- Single unit selected
- Multiple units selected
- All units selected should match "numeric" filter

**review**:
@Workiva/xt
@paulwarren-wk
image


aviary3-wk:
## Security Insights

## Action Items
---

derekgengenbacher-wf (dismissed):

allenbross-wf (dismissed):

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -282,15 +283,37 @@ def addFact(self, f):
 
         if f.isNumeric:
             if f.unit is not None and len(f.unit.measures[0]):
-                # XXX does not support complex units
-                unit = self.nsmap.qname(f.unit.measures[0][0])
-                aspects["u"] = unit
+                # Simplify the complex unit to a single measure by taking the first encountered measure
+                measure = self.nsmap.qname(f.unit.measures[0][0])
+                aspects['m'] = measure
```
brettkail-wk, line 288: Why a separate key for this?

It seems like we should use the [OIM format](https://www.xbrl.org/Specification/oim-common/REC-2021-10-13/oim-common-REC-2021-10-13.html#term-unit-string-representation) for units (Arelle might already have logic for this somewhere).

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -380,6 +403,7 @@ def createViewer(self, scriptUrl=DEFAULT_VIEWER_PATH, useStubViewer=False, showV
         self.taxonomyData["prefixes"] = self.nsmap.prefixmap
         self.taxonomyData["roles"] = self.roleMap.prefixmap
         self.taxonomyData["rels"] = self.getRelationships()
+        self.taxonomyData["units"] = self.unit_map
```
brettkail-wk, line 406: Why do we need more data in the JSON?  Can we build this map in the JavaScript instead?

aaroncameron-wk: The JSON currently only has a simplified measure attribute (by taking the first numerator from the unit). This PR retains that measure attribute but adds the complex form of the unit.

iXBRLViewerPlugin/iXBRLViewer.py
aaroncameron-wk: The `saveLoadableOIM` plugin does [generate a unit string representation](https://github.com/Arelle/Arelle/blob/2e1a74adf20a71389661d0f86c064797ef6a4718/arelle/plugin/saveLoadableOIM.py#L333-L348) per the above spec, the main difference being conditional parentheses. That code isn't easily accessible from this plugin, and I wanted to generate the "readable" label in a similar way anyway so my latest commit recreates the same representation.

For future improvement, I created [this issue](https://github.com/Arelle/Arelle/issues/698) which will allow Arelle and it's plugins to use the same method for unit string representation.

iXBRLViewerPlugin/iXBRLViewer.py
brettkail-wk: Do we still need the first numerator somewhere in ixbrl-viewer?  It seems like we should add full support for units, and then it won't be required any more.

If it is still required, I think the JavaScript should do that, and the serialized format should be complete.  I don't think it's useful to duplicate the unit data in two JSON keys.

iXBRLViewerPlugin/iXBRLViewer.py
brettkail-wk: If Arelle doesn't expose what we need, we can copy the relevant logic into ixbrl-viewer.  I don't think we should do pretty-printing of the local name in Python and save it in the JSON.  We should save the data in a format that the JavaScript can parse and render in whatever way is useful.

iXBRLViewerPlugin/iXBRLViewer.py
aaroncameron-wk: Initially I didn't have a grasp on where all the "measure" was used and wasn't confident that I could remove it cleanly, so opted to add unit alongside it. Reconsidering now though, it doesn't look like it will be very problematic to remove measure entirely so I'll take a shot at it.

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -323,6 +342,27 @@ def addFact(self, f):
         self.taxonomyData["facts"][f.id] = factData
         self.addConcept(f.concept)
 
+    def measureString(self, measure, readable):
+        if readable:
+            return measure.localName.title() if measure.localName.islower() else measure.localName
+        else:
+            return self.nsmap.qname(measure)
+
+    def unitString(self, unit, readable):
```
sagesmith-wf, line 351: Curious why we have to turn this into a string instead of leaving it as a list, I would assume that would be preferred?

brettkail-wk: I had in my head that the JSON was pseudo-OIM.  `"u":[[numers...],[denoms...]]` (e.g., `"u":[["iso4217:USD"]]`).

@paulwarren-wk Do you have thoughts on the design here?

aaroncameron-wk: My intent was that additional information for the unit (a readable label, a monetary flag) be stored in a map somewhere, so having it in string representation aids in using it as a key to said map. I just completed a rework and I've included references to the OIM format in relevant docstrings.

sagesmith-wf (approved):

allenbross-wf (approved):

allenbross-wf:
QA +10

allenbross-wf:
QA +10

https://github.com/orgs/Workiva/teams/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #425, opened 2023-05-17
XT-2876: Scales filter

aaroncameron-wk:
#### Reason for change
Filter facts by scale

#### Description of change
Added scale to generated JSON
Updated scale/accuracy translations
Added display for scale in fact details
Added scale filter

#### Steps to Test

**review**:
@Workiva/xt
@paulwarren-wk


derekgengenbacher-wf (dismissed):

iXBRLViewerPlugin/viewer/src/i18n/en/currencies.json
```diff
@@ -12,10 +12,15 @@
     "centsGBP": "pence",
     "accuracy3": "thousandths",
     "accuracy2": "hundredths",
+    "accuracy1": "tenths",
```
sagesmith-wf, line 15: pre-existing: Naming scheme doesn't seem very intuitive
pre-existing: The inclusion of accuracy in the currencies file doesn't seem very intuitive

Nit: How do we handle decimals?

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -266,31 +266,49 @@ export class Fact {
...
-        if (this.isMonetaryValue()) {
-            var currency = this.report().qname(this.measure()).localname;
-            if (d == 2) {
-                var name = i18next.t(`currencies:cents${currency}`, {defaultValue: name});
-            }
+        const label = this.getAccuracyLabel(d);
+        if (label != null) {
+            return `${d} (${label})`;
+        }
+        else {
```
sagesmith-wf, line 292: Nit: Else isn't needed because the if returns

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -266,31 +266,49 @@ export class Fact {
...
+        if (scale === undefined || scale === null) {
+            return i18next.t("common.unspecified");
         }
-        if (name !== "noName") {
-            d += " ("+name+")";
+        const label = this.getAccuracyLabel(scale, true);
+        if (label != null) {
+            return `${scale} (${label})`;
         }
         else {
```
sagesmith-wf, line 309: pre-existing nit: The else isn't needed here because the if returns

iXBRLViewerPlugin/viewer/src/js/report.js
```diff
@@ -218,6 +219,25 @@ iXBRLReport.prototype.getUnit = function(unitKey) {
...
+    Object.values(this._items).forEach(fact => {
+        const scale = fact.scale();
+        if (scale != null) {
+            if (!(scale in usedScalesMap)) {
+                usedScalesMap[scale] = new Set();
+            }
+            const labels = usedScalesMap[scale];
+            const label = fact.getAccuracyLabel(scale, true);
+            if (label && !labels.has(label)) {
+                labels.add(label);
```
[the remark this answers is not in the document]
paulwarren-wk: Unless I'm misunderstanding, we only show currency for the specific case of hundredths of a monetary value, and I do think there's some benefit in doing so, as filtering for "values reported in cents" seem more intuitive than "values reported in hundredths".

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -262,31 +266,49 @@ export class Fact {
...
         }
-        var name = i18next.t(`currencies:accuracy${d}`, {defaultValue:"noName"});
-        if (this.isMonetaryValue()) {
-            var currency = this.report().qname(this.measure()).localname;
-            if (d == 2) {
-                var name = i18next.t(`currencies:cents${currency}`, {defaultValue: name});
-            }
+        const label = this.getAccuracyLabel(d);
+        if (label != null) {
+            return `${d} (${label})`;
```
paulwarren-wk, line 290: I'd prefer not to include the actual value of the scale attribute, or at least make it less prominent, as I think it's a technical detail, and it's meaning isn't obvious if you're not familiar with iXBRL.

I'd prefer to just say:

> Scale: millions

If it's important to include the value, perhaps do:

> Scale: millions (scale=6)

or just:

> Scale: millions (6)

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -262,31 +266,49 @@ export class Fact {
...
+        if (!this.isNumeric() || this.isNil()) {
+            return i18next.t("common.notApplicable");
+        }
+        const scale = this.scale();
+        if (scale === undefined || scale === null) {
+            return i18next.t("common.unspecified");
         }
-        if (name !== "noName") {
-            d += " ("+name+")";
+        const label = this.getAccuracyLabel(scale, true);
```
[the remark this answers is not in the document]
paulwarren-wk: Only in the specific case of monetary values with `decimals="-2"` and `scale="2"`, and we're effectively saying "accurate to the nearest cent, reported in cents" which seems reasonable.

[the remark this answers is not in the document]
brettkail-wk: Ah, I finally understand now, thanks :-).

There is something confusing about using "accuracy" as the i18n key for describing "scale".  It seems like it might be clearer to invert the keys (accuracy described in terms of scale labels) or to use some other word (exponent?).

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -262,31 +266,49 @@ export class Fact {
         return this.f.err == 'INVALID_IX_VALUE';
     }
 
+    getAccuracyLabel(value, isScale=false) {
+        return this._report.getAccuracyLabel(
+                isScale ? -value : value, // Invert value if this is a scale value
```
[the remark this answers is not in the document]
brettkail-wk: This is now a duplicate of [this thread](https://github.com/Workiva/ixbrl-viewer/pull/425#discussion_r1196865791).

iXBRLViewerPlugin/viewer/src/js/report.js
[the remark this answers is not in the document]
brettkail-wk: Ah, I understand this now, thanks :-).

iXBRLViewerPlugin/viewer/src/i18n/en/currencies.json
aaroncameron-wk: > pre-existing: Naming scheme doesn't seem very intuitive
> pre-existing: The inclusion of accuracy in the currencies file doesn't seem very intuitive

They're used for both scale and accuracy, but I think `scale3` is a more intuitive reference to `thousands` than `accuracy-3`. Maybe making that change and moving them all to a `scale.json` addresses both of these concerns?

> Nit: How do we handle decimals?

Not exactly sure what you mean, but `accuracy1..3` represent 1-to-3 decimal places, respectively. Scale is defined [in spec](https://www.xbrl.org/specification/inlinexbrl-part1/rec-2013-11-18/inlinexbrl-part1-rec-2013-11-18.html#sec-nonFractions) as an integer representing a power of 10, so we don't ever need to worry about `accuracy2.5`, if that's what you're asking about.

iXBRLViewerPlugin/viewer/src/js/fact.js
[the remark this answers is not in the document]
aaroncameron-wk: I [agree](https://github.com/Workiva/ixbrl-viewer/pull/425#discussion_r1198191696) "scale" is the more intuitive label to use over "accuracy". That's what I'll go with unless anyone feels strongly that something more general like `exponent3` or `decimalPlace10^3` is better.

iXBRLViewerPlugin/viewer/src/i18n/en/currencies.json
sagesmith-wf: Will defer to @brettkail-wk @paulwarren-wk here for their xbrl expertise, does ^ seem reasonable?

iXBRLViewerPlugin/viewer/src/js/fact.js
aaroncameron-wk: Agreed. Do you feel the same about the accuracy label?

paulwarren-wk: Yes, I think so.  I think the number is slightly more meaningful for accuracy, as I think "accurate to N decimal places" is more understandable than "scaled by 10 to the N", but on balance I think we should swap it.

Maybe we could do it like this:

> Scale: millions (scale=6)
> Accuracy: millions (decimals=-6)

Which makes the "business" meaning primary, but keeps the XBRL technical detail for those who understand it.  Or possibly we could put the parenthetical bit in a tooltip.

iXBRLViewerPlugin/viewer/src/i18n/en/currencies.json
paulwarren-wk: I agree that "scale" is better than "accuracy".  I was initially confused that `-` was meaning "minus" rather than just being a separator i.e. `accuracy-3` was actually "accuracy = -3", but I don't have a suggestion for how to improve that aspect.

iXBRLViewerPlugin/viewer/src/js/fact.js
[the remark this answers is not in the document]
paulwarren-wk: "scale" WFM.

iXBRLViewerPlugin/viewer/src/js/fact.js
aaroncameron-wk: I like the tooltip to keep clutter low. Perhaps that's something we can incorporate if we implement some kind of abstraction for rending fact details to easily configure tooltips for each detail.

iXBRLViewerPlugin/viewer/src/i18n/en/currencies.json
aaroncameron-wk: See latest commits, ended up creating "scale" object with "-3", "1", "9", etc. as keys.

derekgengenbacher-wf (approved):

allenbross-wf (approved):

derekgengenbacher-wf:

@Workiva/release-management-p

Issue #437, opened 2023-05-31
[BUG] Logic for selecting fact from nested tags doesn't work with continuations


paulwarren-wk:
### What happened?

When you click on nested text block facts in certain reports, the outermost fact is selected rather than the innermost fact.

For example, open [this filing](https://filings.xbrl.org/3TK20IVIUJ8J3ZU0QE75/2022-12-31/ESEF/NL/0/ingbank-2022-12-31-en/reports/ixbrlviewer.html), scroll to page 189, and click on the "Defined benefit plan" text.  You now have two facts  shown in the fact inspector, but it is the outer "Disclosure of notes and other explanatory information" tag that is selected by default, not the innermost one:

![image](https://github.com/Workiva/ixbrl-viewer/assets/45077928/0afaa4bd-3ae1-4a61-af8e-be839e2c552b)

We have some complex logic here:

* For nested tags we generally want to select the innermost tag.
* For "double tags" (i.e. nested tags with exactly the same content) we want to select the outermost tag, as it's more intuitive that the "next tag" button takes you to the second double tag.

This filing uses `ix:continuation` elements, and the contents of the `ix:continuation` elements are the same for both the inner and outer tags, so we mistakenly treat this as a double tag rather than a nested tag.

We are also scrolling the view even though the tag is already in view.  I don't think that we should ever scroll when selecting a fact by clicking on it.

### Version

1.1.61 (ish)

### With which browsers are you experiencing the bug?

Chrome

PR #440, opened 2023-06-05
Never scroll when selecting a fact by clicking on it

paulwarren-wk:
#### Reason for change

Currently, when selecting a fact, we check to see if the fact is fully visible, and if not, scroll so that is is centered in the view.

The code doesn't take into account continuations, so if you click on a continuation, and the "head" fact isn't fully in view, we scroll away from what you clicked on.  This is also a problem even without continuations if you have a large text block that is larger than the viewport.

On reflection, when you select a fact by clicking on it in the document view, I don't think it's ever desirable to scroll that view.

#### Description of change

We now never scroll the view

#### Steps to Test

See steps described on #437 (note this isn't a fix for the main part of #437, just the fact that it scrolls away from what you click on).

Alternatively, scroll so that any fact is only partially visible.  Click on it, and notice that the view no longer scrolls to center it.

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -555,7 +555,9 @@ export class Viewer {
         if (itemId !== null) {
             this.onSelect.fire(itemId, itemIdList, byClick);
         }
-        this.onSelect.fire(null);
+        else {
+            this.onSelect.fire(null);
+        }
     }
 
     // Handle a mouse click to select.  This finds all tagged elements that the
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -547,14 +547,15 @@ export class Viewer {
      *
      * Takes an optional list of factIds corresponding to all facts that a click
      * falls within.  If omitted, it's treated as a click on a non-nested fact.
+     *
+     * byClick indicates that the element was clicked directly, and in this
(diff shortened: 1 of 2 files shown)


iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -547,14 +547,15 @@ export class Viewer {
...
+     *
+     * byClick indicates that the element was clicked directly, and in this
+     * case we never scroll to make it more visible.
      */
-    selectElement(itemId, itemIdList) {
+    selectElement(itemId, itemIdList, byClick) {
         if (itemId !== null) {
-            this.onSelect.fire(itemId, itemIdList);
-        }
-        else {
```
lukepint-wk, line 555: Can you explain the removal of the `else` here? Why do we want to call `this.onSelect.fire(null);` unconditionally?

derekgengenbacher-wf: @paulwarren-wk

paulwarren-wk: This was a mistake.  I somehow imagined a `return` inside the `if`.  Reverted in 13a36c3

brettkail-wk (approved):

lukepint-wk (approved):

derekgengenbacher-wf:

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #455, opened 2023-06-19
Improve layout of tag summary table with narrow width / long prefixes

paulwarren-wk:
#### Reason for change

The new fact summary table does not work well on narrow screen widths or on documents with long prefixes.

For example, this example is from [this Japanese EDINET filing](https://disclosure2.edinet-fsa.go.jp/WEEE0040.aspx?bXVsPWNhbm94JmN0Zj1vZmYmZmxzPW9uJmxwcj1vZmYmcnByPW9mZiZvdGg9b2ZmJnBmcz02Jnllcj0mbW9uPQ==)

image

The long prefix names causes inconsistent column widths for the remaining width, and inconsistent wrapping of the percentage figures.

Problem is also visible with shorter prefixes on a narrow window:

image

#### Description of change

This change fixes the width of the columns containing numbers, and uses diagonal column headers:

image

The percentage figures are forced onto a separate line, and numbers are now right-aligned.

Diagonal column headers aren't perfect, as we have to guess how much vertical space to leave (or use JS to adjust it - I can't see a way to do it in CSS), bearing in mind that i18n might change the length of these labels.

The only other solution I can see is moving this into a popup dialog where we would have more horizontal space.

#### Steps to Test

Take any iXBRL document and reduce the window width (or inspector pane width) and check the behaviour of the Tag Summary table.

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/html/inspector.html
+++ b/iXBRLViewerPlugin/viewer/src/html/inspector.html
@@ -286,11 +286,11 @@ limitations under the License.
                 <table>
                   <thead>
                     <tr>
-                      <th scope="col" data-i18n="inspector.summary.tag.colNamespace">Namespace</th>
-                      <th scope="col" data-i18n="inspector.summary.tag.colPrimaryItems">Primary Items</th>
-                      <th scope="col" data-i18n="inspector.summary.tag.colDimensions">Dimensions</th>
-                      <th scope="col" data-i18n="inspector.summary.tag.colMembers">Members</th>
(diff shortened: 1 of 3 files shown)


austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #461, opened 2023-06-22
XT-2856: Create GitHub workflow to build npm package

### LINKED ISSUE #450: Create GitHub workflow to build npm package

We should publish an ixbrl-viewer package to npm that includes the production JS build of ixbrl-viewer. To automate this we need to create a GitHub workflow that handles versioning (based on the git tag).

aaroncameron-wk:
#### Reason for change
Closes #450 

#### Description of change
Addition of workflow for publishing npm package on `workflow-dispatch`

#### Steps to Test
Won't be a true test until release but a slightly modified version of this workflow resulted in the publishing of this package: https://www.npmjs.com/package/@aaroncameron-wk/ixbrl-viewer/v/1.1.87-7-gd9f38a1

**review**:
@Workiva/xt
@paulwarren-wk


aviary3-wk:
## Security Insights

## Action Items
---

derekgengenbacher-wf (approved):

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/checkout@v3.3.0
+        with:
+          fetch-depth: 0 # npm run prod requires tags, sadly this is the only way to get them.
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ inputs.node_version }}
+      - name: Install dependencies
+        run: npm install
+      - name: Test with npm
+        run: npm run test
```
austinmatherne-wk, line 26: I found it rather unexpected that running tests in the workflow didn't require installing dev dependencies.

Jest, the linters, and webpack should all fall under devDependencies. However, I would also propose that since we're publishing the built and minified JS file to NPM, all of our dependencies should be classified as devDependencies because most users who install the package won't require any of them. Additionally, I believe it's time we committed the package lock file. @paulwarren-wk @aaroncameron-wk what are your thoughts?

package.json
```diff
@@ -1,11 +1,13 @@
 {
   "name": "ixbrl-viewer",
-  "version": "0.0.0",
```
aaroncameron-wk, line 3: Removing this so it isn't misleading to anyone looking at package.json in the git repo. This gets set immediately immediately before publishing, so it does appear in the actual packaged [package.json](https://www.npmjs.com/package/@aaroncameron-wk/ixbrl-viewer/v/1.1.87-7-gd9f38a1?activeTab=code)

package.json
```diff
@@ -1,11 +1,13 @@
...
   "description": "iXBRL Viewer",
   "main": "iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js",
   "directories": {
     "test": "tests"
   },
+  "files": [
+    "iXBRLViewerPlugin/viewer"
+  ],
```
aaroncameron-wk, line 10: @austinmatherne-wk @paulwarren-wk 
I wouldn't think anything outside of the `/viewer` folder would need to be included in the package (besides the package config/readme included by default). I didn't go down to `/dist` because others may want to consume the `/src` JS, which w-ixbrl-viewer currently does via import (which uses the `main` attribute above)

austinmatherne-wk: I think this is reasonable for the future, but I suspect that right now it will break the w-ixbrl-viewer build as it depends on pulling in the Python code through the NPM install

package.json
austinmatherne-wk: Does this still work with local installs? `npm install ../ixbrl-viewer`?

.github/workflows/npm-publish.yml
austinmatherne-wk: Good catch 👍

I think all of the build and test dependencies can still go under dev, but yes, leave the imported dependencies in prod.

package.json
aaroncameron-wk: `install` does, but `pack` doesn't, and I missed that the Dockerfile was already doing a replace on the `0.0.0` version string so I reverted to that. I just didn't love that the version string would always show as `0.0.0` to anyone checking out the repo, but I'm fine with it for now.

derekgengenbacher-wf (approved):

package.json
```diff
@@ -30,40 +30,40 @@
...
   "dependencies": {
-    "babel-core": "^6.26.3",
-    "babel-jest": "^26.0.1",
-    "babel-preset-env": "^1.7.0",
-    "base64-inline-loader": "^2.0.1",
-    "i18next": "^20.3.2",
-    "jquery-i18next": "^1.2.1",
     "chart.js": "^2.9.3",
     "core-js": "^3.6.4",
     "css-loader": "^3.4.2",
```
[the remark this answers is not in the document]
aaroncameron-wk: I thought so initially as well, but it appears that `css-loader` and `less-loader` are `require`-ed in a non-traditional way [here](https://github.com/Workiva/ixbrl-viewer/blob/0b03bd1b18b46c4c30b59e40de47332d33804022/iXBRLViewerPlugin/viewer/src/js/viewer.js#L410). Although to be honest I'm not really sure what that code is doing, but I figured it meant that less was being compiled and loaded here.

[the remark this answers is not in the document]
austinmatherne-wk: Good catch

austinmatherne-wk (approved):

derekgengenbacher-wf:

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #467, opened 2023-06-26
Modernize JavaScript in inspector.js

### LINKED ISSUE #377: [REQUEST] Upgrade to ES6 syntax

We should upgrade to ES6 syntax in attempts to modernize the app

paulwarren-wk:
#### Reason for change

Fixes #377.

#### Description of change

Updates `inspector.js` to use ES `class` syntax, and other more modern JS constructs.  Reduces use of jQuery.

Also fixes some minor issues in the updates to `viewer.js`.

#### Steps to Test

Viewer functionality should be unchanged.  Check that inspector operations work correctly as before.

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/inspector.js
+++ b/iXBRLViewerPlugin/viewer/src/js/inspector.js
@@ -820,22 +820,9 @@ export class Inspector {
     }
 
     _updateEntityIdentifier(fact, context) {
-        const cell = $('tr.entity-identifier td', context);
-        cell.empty();
-        const url = Identifiers.identifierURLForFact(fact);
-        const name = Identifiers.identifierNameForFact(fact);
-        if (name !== undefined) {
-            $('<span></span>').text('[' + name + "] ").appendTo(cell)
-            if (url !== undefined) {
-                $('<a target="_blank"></a>').attr('href',url).text(fact.identifier().localname).appendTo(cell)
-            }
-            else {
-                $('<span></span>').text(fact.identifier().localname).appendTo(cell);
-            }
-        }
-        else {
-            cell.text(fact.f.a.e);
-        }
+        $('tr.entity-identifier td', context)
+            .empty()
+            .append(Identifiers.readableNameHTML(fact.identifier()));
     }
 
     _footnoteFactsHTML(fact) {
--- a/iXBRLViewerPlugin/viewer/src/js/inspector.js
+++ b/iXBRLViewerPlugin/viewer/src/js/inspector.js
@@ -573,7 +573,6 @@ export class Inspector {
         }
     }
 
-
     _anchorList(fact, anchors) {
         const html = $("<ul></ul>");
         if (anchors.length > 0) {
@@ -827,7 +826,7 @@ export class Inspector {
 
     _footnoteFactsHTML(fact) {
         const html = $('<div></div>');
-        fact.linkedFacts.forEach((linkedFact) =>  {
+        fact.linkedFacts.forEach((linkedFact) => {
             html.append(this.factListRow(linkedFact));
         });
         return html;
@@ -931,8 +930,7 @@ export class Inspector {
     }
 
     update() {
-        const inspector = this;
-        const cf = inspector._currentItem;
+        const cf = this._currentItem;
         if (!cf) {
             $('#inspector').removeClass('footnote-mode');
             $('#inspector').addClass('no-fact-selected');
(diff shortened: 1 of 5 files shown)


aviary-wf:
## Security Insights
<details>
<li>Watched keyword <code>.html(s</code> in <code>iXBRLViewerPlugin/viewer/src/js/inspector.js</code> line(s) ['770'] added</li>
</details>

## Action Items
---

austinmatherne-wk (approved):

lukepint-wk (approved):

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -34,1067 +34,1061 @@ import { DIMENSIONS_KEY, DocumentSummary, MEMBERS_KEY, PRIMARY_ITEMS_KEY, TOTAL_
...
-    $('.calculations .item').filter(function () {   
-        return $.inArray(id, $.map($(this).data('ivid'), f => f.id )) > -1 
-    }).addClass('linked-highlight');
-    $('#inspector .search .results tr').filter(function () {   
-        return $(this).data('ivid') == id;
-    }).addClass('linked-highlight');
-}
+    viewerMouseEnter(id) {
+        $('.calculations .item')
+            .filter((i, e) => Object.keys($(e).data('ivid') ?? {}).includes(id))
```
brettkail-wk, line 703: Can `Object.keys(o).includes(k)` be simplified to `k in o`?

To be sure, this expression doesn't seem equivalent to the other one, which seemed to be calling `f.id` on each entry of `ivid` (so it's an array of objects with `id` attributes).  Is this change fixing or introducing a bug?  I'm getting lost trying to evaluate `resolvedCalculation`, so I'm not sure if `itemHTML.data('ivid', r.facts);` is setting a list of objects or ids.  Perhaps some more doc there would help.

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/inspector.js
paulwarren-wk: > Can `Object.keys(o).includes(k)` be simplified to `k in o`?

Yes, that would be better.

> To be sure, this expression doesn't seem equivalent to the other one, which seemed to be calling `f.id` on each entry of `ivid` (so it's an array of objects with `id` attributes). Is this change fixing or introducing a bug? 

It's not doing the same thing, but should have the same end result.  `r.facts` is a map of fact IDs to fact objects, so the key is the same as the value's `id` property.  I'm really struggling to understand what I was thinking when I wrote the original, as the only reason I can think of for using this data structure rather than an array would have been do an efficient check if a given ID is present.

> I'm getting lost trying to evaluate `resolvedCalculation`, so I'm not sure if `itemHTML.data('ivid', r.facts);` is setting a list of objects or ids. Perhaps some more doc there would help.

Yes - I'll add some further comments.  I don't think we ever actually use the object values of the `ivid` property, so this could just be a list of IDs, and given a better name (e.g. `ivids`) as we use `ivid` to hold a single ID elsewhere.

I've got another long-standing branch for Calc 1.1 which already rewrites a lot of this code, so for this PR I'll just add some comments to better document the current behaviour.

brettkail-wk: I did not realize that `$.map(object, f)` worked on object properties, I thought it was only for arrays.  Now I think I understand how it's working.  But I think `ivid` is always set to `list<string>` elsewhere, so it's confusing that it's set to `map<string,object>` for these calcs.

paulwarren-wk: Agreed, it was very confusing.

I've now changed this usage to be a list, and renamed it to `ivids`.

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -34,1067 +34,1061 @@ import { DIMENSIONS_KEY, DocumentSummary, MEMBERS_KEY, PRIMARY_ITEMS_KEY, TOTAL_
...
-    if (fact !== undefined) {
-        $('<span></span>').text('['+Identifiers.identifierNameForFact(fact) + "] ").appendTo(cell)
-        if (url !== undefined) {
-            $('<a target="_blank"></a>').attr('href',url).text(fact.identifier().localname).appendTo(cell)
+    _updateEntityIdentifier(fact, context) {
+        const cell = $('tr.entity-identifier td', context);
+        cell.empty();
+        const url = Identifiers.identifierURLForFact(fact);
+        const name = Identifiers.identifierNameForFact(fact);
+        if (name !== undefined) {
```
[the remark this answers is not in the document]
paulwarren-wk: Ah yes, I spotted this when doing the refactor.

I've now further refactored and added tests.

brettkail-wk (approved):

austinmatherne-wk (approved):

derekgengenbacher-wf (approved):
Security +1

rmconsole-wf (approved):
+1 from RM

PR #446, opened 2023-06-14
XT-2997: Handle null unit in scale filter

austinmatherne-wk:
#### Reason for change
The scale filter currently throws if a fact unit is null.

Screenshot 2023-06-14 at 11 45 17 AM

#### Description of change
Don't pass an undefined measure to `this.report().qname(...)`.

#### Steps to Test
1. `npm run prod`
2. Use [filing_documents.zip](https://github.com/Workiva/ixbrl-viewer/files/11748087/filing_documents.zip)
3. `python arelleCmdLine.py --plugins=/path/to/ixbrl-viewer/iXBRLViewerPlugin -f /path/to/filing_documents.zip --save-viewer ixbrl-report-viewer.html --viewer-url=/path/to/ixbrl-viewer/iXBRLViewerPlugin/viewer/dist/ixbrlviewer.js`
4. Open `ixbrl-report-viewer.html` and click into the fact and confirm that an exception isn't thrown.

**review**:
@Workiva/xt
@paulwarren-wk

PR #476, opened 2023-06-29
Additional test for calling getScaleLabel on facts with no units.

### LINKED ISSUE #435: [BUG] getUsedScalesMap fails on reports with missing taxonomy definitions
### What happened?

If you open a report containing numeric concepts that are missing taxonomy definitions (e.g. because the relevant taxonomy file could not be found), then you get an error on loading the document:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'split')
    at new QName (qname.js:17:1)
    at iXBRLReport.qname (report.js:248:1)
    at Fact.getScaleLabel (fact.js:284:1)
    at report.js:234:40
    at Array.forEach (<anonymous>)
    at iXBRLReport.getUsedScalesMap (report.js:227:1)
    at Inspector._getScalesOptions (inspector.js:344:1)
    at Inspector.setupSearchControls (inspector.js:333:1)
    at Inspector.searchReady (inspector.js:372:1)
    at inspector.js:154:17
```

The search controls are stuck with the "Please wait - building index..." mask.

The problem is that we don't include the units for these facts in the JSON, because, without the type information we don't know that they're numeric.  `getUsedScalesMap` reasonably enough assumes that if they have a scale, that they're numeric and calls `getScaleLabel` which fails because it can't find the unit.

Obviously there are some simple fixes for this specific issue, but I do have a general concern with the approach of continuing in the face of serious XBRL validity errors, as even with this bug fixed, there will be lots of things that don't work in the viewer.   I think that Arelle should make it much more obvious when you're working with a report that is not valid according to the specifications, as currently, serious validity errors are mixed in with messages relating to business-level errors from formula rules.

### Version

1.1.76-dev

Chrome

### Documents

_No response_

### Screenshots

_No response_

paulwarren-wk:
#### Reason for change

Additional tests for the fix on #446 that more closely reflect what caused #435.  The underlying issue on #435 was calling `getScaleLabel` on a fact with absent units, rather than `null` units. 

`getScaleLabel` was previously tested via `readableAccuracy` which has its own guard for non-numeric facts, but it's also called directly from `getUsedScaleMap` on anything with a `scale` attribute.

#### Description of change

Additional tests that call `getScaleLabel` with missing units.

Closes #435 .

#### Steps to Test

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/fact.test.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.test.js
@@ -441,6 +441,34 @@ describe("Readable accuracy", () => {
     });
 });
 
+describe("Readable accuracy", () => {
+    test("With units", () => {
+        expect(testFact({
+            "v": "1234",
+            "d": 4,
+            "a": { "u": "iso4217:GBP" }
+        }).getScaleLabel(-2)).toBe("pence");
+        expect(testFact({
+            "v": "1234",
+            "d": 4,
+            "a": { "u": "iso4217:GBP" }
+        }).getScaleLabel(-4)).toBe(null);
+    });
+    test("Without units", () => {
+        expect(testFact({
+            "v": "1234",
+            "d": 4,
+            "a": {}
+        }).getScaleLabel(-2)).toBe("hundredths");
+        expect(testFact({
+            "v": "1234",
+            "d": 4,
+            "a": {}
+        }).getScaleLabel(-4)).toBe(null);
+    });
+});
+
+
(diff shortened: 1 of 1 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

austinmatherne-wk (approved):

brettkail-wk (approved):

rmconsole-wf (approved):
+1 from RM

PR #518, opened 2023-07-13
XT-3013: Create ixbrl-viewer review mode

### LINKED ISSUE #453: Add review mode with untagged numbers highlighting

We should add the ability to highlight untagged numbers and dates similar the XBRL tag highlighting that currently exists.

aaroncameron-wk:
#### Reason for change
Closes #453 

#### Description of change

- Added `reviewMode` to `iXBRLViewer.options`
  - Populated from query parameter
  - Can be used to conditionally perform review-mode specific tasks
- ~Added text to banner when review mode is enabled~
- When review mode is enabled:
  - Preprocessing for untagged elements occurs
  - Additional options are added to the toolbar
  - Highlighting key changes

Note: `number-matcher*` files were brought over from Workiva viewer and changed as little as possible to conform to `ixbrl-viewer` environment without making functional changes. 

Screenshot 2023-07-18 at 12 48 23 PM

#### Steps to Test

- Generate three viewers:
  - A: One using `master` to demonstrate current default behavior
  - B: One using Workiva's viewer to demonstrate intended review mode behavior
  - C: One using this branch
 - Compare A to C without review mode, confirm highlighting looks the same
 - Compare B to C with review mode enabled, confirm highlighting looks the same

**review**:
@Workiva/xt
@paulwarren-wk


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -723,7 +799,8 @@ export class Viewer {
     }
 
     _setTitle(docIndex) {
-        $('#top-bar .document-title').text($('head title', this._iframes.eq(docIndex).contents()).text());
+        const mode = this._iv.options.reviewMode ? " - Review Mode" : "";
```
austinmatherne-wk, line 802: We may want to add this in the future (and even a button to switch between modes), but for now we should leave the title unchanged

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -264,6 +268,7 @@ iXBRLViewer.prototype.setProgress = function (msg) {
          * https://bugs.chromium.org/p/chromium/issues/detail?id=675795 
          */
         window.requestAnimationFrame(function () {
+            console.log(`%c [Progress] ${msg} `, 'background: #77d1c8; color: black;');
```
paulwarren-wk, line 271: Is this console logging intentional?

iXBRLViewerPlugin/viewer/src/less/viewer.less
```diff
@@ -141,3 +141,23 @@ div.ixbrl-table-handle {
...
+.review.review-highlight-untagged-numbers .review-untagged-number {
+  background-color: @highlight-1;
+}
+
+.review.review-highlight-untagged-dates .review-untagged-date {
+  background-color: @highlight-2;
+}
+
+.review .ixbrl-highlight:not(.ixbrl-no-highlight) {
+  background-color: @highlight-default !important;
```
paulwarren-wk, line 154: Now that this is all in one file, it'd be nice to be able to lose the `!important`s on these lines, which I think we could do by adding a `:not(.review)` block where the highlighting is originally applied around line 11, and then dropping these overrides altogether.

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
aaroncameron-wk: I found it helpful for my own use when learning about the pre-processing pipeline but if we're trying to keep the dev console clean I'm happy to exclude this. 

Maybe a `?dev` flag is something we could consider?

iXBRLViewerPlugin/viewer/src/less/viewer.less
aaroncameron-wk: Actually I think because `.review .ixbrl-highlight:not(.ixbrl-no-highlight)` is narrower than `.ixbrl-highlight:not(.ixbrl-no-highlight)` it automatically supersedes the default. Though I'm not 100% confident in my understanding, it appears to have the desired end result when I simply exclude `!important`.

austinmatherne-wk (approved):

iXBRLViewerPlugin/viewer/src/less/viewer.less
```diff
@@ -6,14 +6,14 @@
 @import (reference) "icons.less";
 
 .ixbrl-highlight:not(.ixbrl-no-highlight) {
-  background-color: @highlight-default !important;
+  background-color: @highlight-default;
```
austinmatherne-wk, line 9: Unfortunately this doesn't work for values within tables. The alternate table row highlighting is applied with inline styles and takes a higher precedence.

iXBRLViewerPlugin/viewer/src/less/viewer.less
austinmatherne-wk: Unfortunately this doesn't work for values within tables without `!important`. The alternate table row highlighting is applied with inline styles and takes a higher precedence

paulwarren-wk: Yes, sorry, I was misremembering what this `!important` was for.  As you say, it's to override explicit background colors in the source document, rather than to override base styles from the plugin.

I think it would make sense to re-order the file so that these styles are immediately after the non-review styles.

austinmatherne-wk: It would also be good to add a comment explaining why the `!important`s are there

austinmatherne-wk (approved):

austinmatherne-wk:
# Verified

* review mode

QA +10
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #545, opened 2023-07-27
XT-3096: Integrate viewer url and feature options into the tools menu

### LINKED ISSUE #520: Integrate viewer url and mode options into the tools menu

At present, the JavaScript viewer-url can only be defined [via the command line](https://github.com/Workiva/ixbrl-viewer/blob/acf9bedbe75d08e12fea3bb55c8b988faa95e72c/iXBRLViewerPlugin/__init__.py#L47-L51) (as referenced in #373) or through the [save ixbrl viewer instance dialog](https://github.com/Workiva/ixbrl-viewer/blob/acf9bedbe75d08e12fea3bb55c8b988faa95e72c/iXBRLViewerPlugin/ui.py#L25).

#519 introduces another command-line option for setting the viewer mode, which should also be adjustable via the GUI.

These two options need to be integrated [into the tools menu](https://github.com/Workiva/ixbrl-viewer/blob/acf9bedbe75d08e12fea3bb55c8b988faa95e72c/iXBRLViewerPlugin/__init__.py#L125-L129) as persistent configurable options. In line with other configuration options in the Arelle GUI, they should not interfere with the settings of the CLI.

aaroncameron-wk:
*Recommend reviewing by commit!*

#### Reason for change
Closes #520 

#### Description of change
- Implemented `FeatureConfig` which can be used to easily add additional features going forward
- Fixed "Load on launch" setting which hasn't been working since [this commit](https://github.com/Arelle/w-ixbrl-viewer). The setting was not being checked causing Arelle to *always* launch on load.
- Split save dialog into two separate dialogs:
  - Save dialog: for saving a viewer instance, initialized with default values from config but doesn't overwrite them
  - Settings dialog: for saving default settings that populate the save dialog and are used by launch-on-load
- Moved "launch-on-load" toggle from View menu to new Settings dialog
- Configured launch-on-load to generate viewer using configured settings (features and script URL)

### Updated Menu:
Screenshot 2023-07-27 at 12 07 10 PM

### Save Dialog:
Screenshot 2023-08-01 at 11 44 01 AM

### Settings Dialog
Screenshot 2023-08-01 at 11 43 14 AM

#### Steps to Test

- Generate viewers with various combinations of: 
  - Both pathways: Load-on-launch, `Save Viewer...`
  - Different settings: Review Mode on/off
  - Different viewer URLs
- Confirm settings saved in settings dialog interact as expected with settings that populate in save dialog, and vice versa
- Confirm viewer does *not* launch when loading a document if launch-on-load is disabled
- Confirm `Reset Defaults` button on settings dialog works as expected

**review**:
@Workiva/xt
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/__init__.py
+++ b/iXBRLViewerPlugin/__init__.py
@@ -253,6 +253,9 @@ class iXBRLViewerLocalViewer(LocalViewer):
 
 def guiRun(cntlr, modelXbrl, attach, *args, **kwargs):
     """ run iXBRL Viewer using GUI interactions for a single instance or testcases """
+    if not cntlr.config.setdefault(CONFIG_LAUNCH_ON_LOAD, DEFAULT_LAUNCH_ON_LOAD):
+        # Don't run on launch if the option has been disabled
+        return
     try:
         import webbrowser
         global tempViewer
--- a/iXBRLViewerPlugin/__init__.py
+++ b/iXBRLViewerPlugin/__init__.py
@@ -16,8 +16,10 @@ from arelle.LocalViewer import LocalViewer
 from arelle.ModelDocument import Type
 from arelle.webserver.bottle import static_file
 
-from iXBRLViewerPlugin.constants import DEFAULT_VIEWER_PATH
-from .iXBRLViewer import IXBRLViewerBuilder, IXBRLViewerBuilderError, VIEWER_FEATURES_AND_DESCRIPTIONS
+from .constants import CONFIG_FEATURE_PREFIX, CONFIG_LAUNCH_ON_LOAD, \
+    CONFIG_SCRIPT_URL, DEFAULT_LAUNCH_ON_LOAD, DEFAULT_OUTPUT_NAME, \
+    DEFAULT_VIEWER_PATH, FEATURE_CONFIGS
+from .iXBRLViewer import IXBRLViewerBuilder, IXBRLViewerBuilderError
 
 
 #
@@ -77,9 +79,9 @@ def iXBRLViewerCommandLineOptionExtender(parser, *args, **kwargs):
                       help="Converts the viewer output into a self contained zip")
     featureGroup = OptionGroup(parser, "Viewer Features",
                             "See viewer README for information on enabling/disabling features.")
-    for featureName, featureDescription in VIEWER_FEATURES_AND_DESCRIPTIONS.items():
-        arg = f'--viewer-feature-{featureName}'
-        featureGroup.add_option(arg, arg.lower(), action="store_true", default=False, help=featureDescription)
+    for featureConfig in FEATURE_CONFIGS:
+        arg = f'--viewer-feature-{featureConfig.key}'
+        featureGroup.add_option(arg, arg.lower(), action="store_true", default=False, help=featureConfig.description)
     parser.add_option_group(featureGroup)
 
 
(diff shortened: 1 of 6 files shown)


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/ui.py
```diff
@@ -107,3 +178,39 @@ def filename(self):
...
+
+        self._launchOnLoad = BooleanVar()
+        self._launchOnLoad.set(self.cntlr.config.setdefault(CONFIG_LAUNCH_ON_LOAD, True))
+
+        super(SettingsDialog, self).__init__(cntlr, "iXBRL Viewer Settings")
+
+    def addFields(self, frame: Frame, y: int) -> int:
+        y += 1
+        launchOnLoadCheckbutton = Checkbutton(frame, text="Launch on Load", variable=self._launchOnLoad, onvalue=True, offvalue=False)
+        launchOnLoadLabel = Label(frame, text="Launches an instance of the viewer in Stub Viewer Mode whenever a document is loaded.")
```
paulwarren-wk, line 200:
```suggestion
        launchOnLoadLabel = Label(frame, text="Launches an instance of the viewer in your default browser  whenever an Inline XBRL report is loaded.")
```

I think that the fact that the local viewer uses stub viewer mode is incidental here.

iXBRLViewerPlugin/ui.py
```diff
@@ -72,33 +80,96 @@ def __init__(self, cntlr):
...
+        self.cntlr = cntlr
+        self.parent = cntlr.parent
+
+        self._filename = StringVar()
+        self._filename.set(self.cntlr.config.setdefault(CONFIG_OUTPUT_FILE, ""))
+        self._zipViewerOutput = BooleanVar()
+        self._zipViewerOutput.set(self.cntlr.config.setdefault(CONFIG_ZIP_OUTPUT, False))
+        self.accepted = False
+
+        super(SaveViewerDialog, self).__init__(cntlr, "Save iXBRL Viewer")
```
austinmatherne-wk, line 130: Typically you call super before any customizations in the sub class. Does something break if you move this to the first line of the `__init__`?

iXBRLViewerPlugin/__init__.py
```diff
@@ -251,12 +255,26 @@ def getLocalFile(self, file, relpath, request):
...
+    if not cntlr.config.setdefault(CONFIG_LAUNCH_ON_LOAD, True):
+        # Don't run on launch if the option has been disabled
+        return
     try:
         import webbrowser
         global tempViewer
         tempViewer = tempfile.TemporaryDirectory()
-        viewer_file_name = 'ixbrlviewer.html'
-        generateViewer(cntlr, tempViewer.name, useStubViewer=True)
+        viewer_file_name = DEFAULT_OUTPUT_NAME
```
austinmatherne-wk, line 265: This is missing an import for `DEFAULT_OUTPUT_NAME`

iXBRLViewerPlugin/ui.py
```diff
@@ -72,33 +80,96 @@ def __init__(self, cntlr):
...
-        self.cntlr.config["iXBRLViewerFileDir"] = os.path.dirname(self._filename.get())
+        """
+        Saves output-specific config values before triggering instance to be saved.
+        :param event:
+        :return:
+        """
+        # Save output-specific configs for next time
+        self.cntlr.config[CONFIG_FILE_DIRECTORY] = os.path.dirname(self._filename.get())
+        self.cntlr.config[CONFIG_OUTPUT_FILE] = self._filename.get()
+        self.cntlr.config[CONFIG_ZIP_OUTPUT] = self._zipViewerOutput.get()
```
austinmatherne-wk, line 154: An option to somehow reset ixbrl-viewer settings to their defaults would be nice.

iXBRLViewerPlugin/ui.py
aaroncameron-wk: Yes, the bulk of the UI layout/logic happens in `ViewerDialog.__init__`, and it depends on these variables being initialized ahead of time. I wouldn't have done it this way if it wasn't an established pattern both here and in Arelle dialogs. I think a better solution is to move the rendering logic out of the initialization code and move it to some `render()` function.

austinmatherne-wk: Thanks for the explanation 🙂

iXBRLViewerPlugin/ui.py
aaroncameron-wk: Added `Reset Defaults` dialog button that resets the dialog variable values, which can then be saved normally by pressing `OK`. I preferred to do it this way rather than resetting, saving, and closing the dialog with one click because it allows the user to view the defaults and cancel or make changes before committing to the defaults.

austinmatherne-wk (approved):

PR #557, opened 2023-09-03
Build font file in CI from SVG icons

paulwarren-wk:
#### Reason for change

The viewer uses a custom font file for icons.  This has previously been built by extracting icons from other Workiva font files, building the font using fontforge and then committing the result (but not the source font files).  This is obviously not ideal as it makes it difficult for anyone else to maintain the font.

#### Description of change

The icons used by the viewer are now committed to the repository as individual SVG files, and are built into a font file using the `webfont` node package as part of the build process.

#### Steps to Test

Check that icons appear correctly in the viewer.

**review**:
@Arelle/arelle
@paulwarren-wk


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/less/icons.less.njk
```diff
@@ -0,0 +1,13 @@
+// See COPYRIGHT.md for copyright information
+
+.viewericon(@char) {
+  font-family: viewericons;
+  content: @char;
+}
+
+{% for glyph in glyphs %}
+.icon-{{ glyph.name }} {
+  .viewericon("\{{ glyph.unicode[0].charCodeAt(0).toString(16) }}");
```
brettkail-wk, line 10: Out of curiosity, what is in the `unicode` array.  I tried searching for documentation about template variables in the webfont project, but [this](https://www.npmjs.com/package/webfont#user-content-template) was all I found and didn't have much detail (assuming that's the right thing).

The `charCodeAt(0)` seems surprising.  If it were a two-code-unit string, it seems like it should be `\uxxxx\uxxxx` rather than just taking the first code unit.  Does the system somehow prevent that?

The `toString(16)` seems surprising.  If the code unit has fewer than 4 hex digits, it seems like it would produce malformed CSS.  Does the system somehow prevent that?

paulwarren-wk: I have to confess that this is a straight copy-paste from the [standard CSS template](https://github.com/itgalaxy/webfont/blob/master/templates/template.css.njk#L134) and I didn't even look at what it was doing.

From a bit of fiddling, it seems that `unicode` is a pair of (unicode character, glyph name) although curiously glyph name  has had `-` replaced with `_`.

I think the next two bits rely on the fact that `webfont` allocates characters starting at `\uEA01` and that private use area stops at `\uF8FF` so it's "always" going to be a 4-digit hex string.

austinmatherne-wk (dismissed):

austinmatherne-wk (approved):
I pushed some commits to this branch for documentation, gitignore, and [workflow compatibility with Arelle](https://github.com/Arelle/Arelle/pull/857).

PR #567, opened 2023-10-02
IXDS bug fixes

austinmatherne-wk:
#### Reason for change
IXDS viewer currently includes all referenced docs, including schemas. At runtime this generates tabs for all of the inline docs and referenced schemas (this can also cause an exception to be raised depending on how the viewer is hosted):
Screenshot 2023-10-01 at 10 41 20 PM

Additionally, the fake "_IXDS" doc is included as a local doc. This causes it to be included in the file summary:
Screenshot 2023-10-01 at 10 41 35 PM

#### Description of change
* Only include referenced inline docs in docset.
* Filter out fake IXDS doc from local docs.

#### Steps to Test
1. `mkdir viewer-out`
1. `python arelleCmdLine.py --plugins 'iXBRLViewerPlugin|inlineXbrlDocumentSet|transforms/SEC' --file '[{"ixds":[{"file":"filing_documents.zip"}]}]' --save-viewer viewer-out`
1. `python -m http.server --directory viewer-out`
1. open http://localhost:8000/avkr-20230630.htm and confirm tabs for `avkr-20230630.htm` and `avkr-20230630_d2.htm` both work.
1. `mkdir viewer-stub-out`
1. `python arelleCmdLine.py --plugins 'iXBRLViewerPlugin|inlineXbrlDocumentSet|transforms/SEC' --file '[{"ixds":[{"file":"filing_documents.zip"}]}]' --save-viewer viewer-stub-out --use-stub-viewer`
1. `python -m http.server --directory viewer-stub-out`
1. open http://localhost:8000/ixbrlviewer.html and confirm tabs for `avkr-20230630.htm` and `avkr-20230630_d2.htm` both work.
1. Open zip in GUI and select both `avkr-20230630.htm` and `avkr-20230630_d2.htm` with the three plugins enabled and confirm the same.

[filing_documents.zip](https://github.com/Arelle/ixbrl-viewer/files/12779281/filing_documents.zip)

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -412,6 +412,7 @@ class IXBRLViewerBuilder:
             xmlDocsByFilename = {
                 os.path.basename(self.outputFilename(doc.filepath)): deepcopy(doc.xmlDocument)
                 for doc in sorted(dts.modelDocument.referencesDocument.keys(), key=lambda x: x.objectIndex)
+                if doc.type == Type.INLINEXBRL
             }
             docSetFiles = list(xmlDocsByFilename.keys())
 
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -445,9 +445,9 @@ class IXBRLViewerBuilder:
 
         localDocs = defaultdict(set)
         for path, doc in dts.urlDocs.items():
-            if isHttpUrl(path):
+            if isHttpUrl(path) or doc.type == Type.INLINEXBRLDOCUMENTSET:
                 continue
-            if doc.type in (Type.INLINEXBRL, Type.INLINEXBRLDOCUMENTSET):
+            if doc.type == Type.INLINEXBRL:
                 localDocs[doc.basename].add('inline')
             elif doc.type == Type.SCHEMA:
                 localDocs[doc.basename].add('schema')
--- a/tests/unit_tests/iXBRLViewerPlugin/test_iXBRLViewer.py
+++ b/tests/unit_tests/iXBRLViewerPlugin/test_iXBRLViewer.py
@@ -329,13 +329,21 @@ class TestIXBRLViewer(unittest.TestCase):
                 Mock(
                     xmlDocument=etree.ElementTree(root),
                     filepath='a.xml',
-                    objectIndex=0
+                    objectIndex=0,
+                    type=Type.INLINEXBRL,
                 ): [],
                 Mock(
                     xmlDocument=etree.ElementTree(root),
                     filepath='b.xml',
-                    objectIndex=1
-                ): []
+                    objectIndex=1,
+                    type=Type.INLINEXBRL,
+                ): [],
+                Mock(
+                    xmlDocument=etree.ElementTree(root),
+                    filepath='a.xsd',
+                    objectIndex=2,
+                    type=Type.SCHEMA,
+                ): [],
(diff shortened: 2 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -412,6 +412,7 @@ def createViewer(self, scriptUrl: str = DEFAULT_VIEWER_PATH, useStubViewer: bool
             xmlDocsByFilename = {
                 os.path.basename(self.outputFilename(doc.filepath)): deepcopy(doc.xmlDocument)
                 for doc in sorted(dts.modelDocument.referencesDocument.keys(), key=lambda x: x.objectIndex)
+                if doc.type == Type.INLINEXBRL
```
brettkail-wk, line 415: > IXDS viewer currently includes all referenced docs, including schemas

If we don't include the iXBRL+linkbases, does this make the viewer no longer valid iXBRL?  I remember [this discussion](https://github.com/Arelle/ixbrl-viewer/pull/31#discussion_r287838780) from long ago.  We might have given up on that in a subsequent PR, but I've forgotten.

Perhaps it's separately useful to create a small architecture doc somewhere that documents goals and non-goals?  #379 and #386 also had some good discussions about our attempt maintain HTML+XHTML subset compatibility (non-goal of fixing input documents not written to that subset), and maintaining security posture of the input document (particularly for extracted text blocks: if input contains literal `<script>`, then extracted text blocks can too, but if input does not, then extracted text blocks must not).

derekgengenbacher-wf (approved):

iXBRLViewerPlugin/iXBRLViewer.py
paulwarren-wk: I think the problem that this is addressing a straightforward bug.  `iXBRLViewer.addfile()` is supposed to be for Inline XBRL files, and they'll get re-serialized using `XHTMLSerializer`, so it wasn't appropriate for schemas and linkbases to be included here.

For non-document set inputs, taxonomies are not included at all.

I'd be interested to know when this bug was introduced, as it certainly used to be possible to have a document set that didn't result in tabs for the taxonomy files.

I think it would be good to have an architecture document to address the points that you mention.  I also wonder whether we should reconsider the requirement for Inline XBRL validity of the output.

austinmatherne-wk: I can't speak to that conversation or intent, but we already don't bundle local linkbases, only documents directly referenced by the iXBRL doc (we do provide a download for the original filing if the source is a zip, but that's separate).

@paulwarren-wk?

paulwarren-wk: > I'd be interested to know when this bug was introduced, as it certainly used to be possible to have a document set that didn't result in tabs for the taxonomy files.

I now see that this is a result of how Arelle is being invoked.

```
--file '[{"ixds":[{"file":"filing_documents.zip"}]}]'
```
If I'm understanding correctly, this is taking all the documents in the ZIP (including schemas and linkbases) and treating them as a document set, which seems a bit dubious, as only Inline XBRL documents can be part of a document set.  I wonder if this `ixds` notation should only look for HTML/XHTML in the ZIP, make a document set out of them, and then include the other files via DTS discovery?

That said, the fix on this PR still makes sense.

paulwarren-wk (approved):

allenbross-wf (approved):

Issue #571, opened 2023-10-11
[REQUEST] - Add a "viewer" url parameter 


lukepint-wk:
### What should we change and why?

Add a new "viewer" url parameter to determine if the Inline Viewer experience should execute on page load.

This would allow hosting a single copy of the report with an easy way of disabling the viewer.

paulwarren-wk:
This would be straightforward to add, although I'd note that if you're using stub viewer mode, you can already load the non-viewer document directly.

I think this option probably only makes sense for non-stub-viewer mode anyway.

paulwarren-wk:
As an example of how stub viewer mode could help here, when using stub viewer mode, the viewer entry point is `ixbrlviewer.html`, e.g.:

https://filings.xbrl.org/213800DAZ89VABJPBD93/2023-06-30/ESEF/DK/0/213800DAZ89VABJPBD93-2023-06-30-da/reports/ixbrlviewer.html

This loads the original (almost) unmodified filing from here:

https://filings.xbrl.org/213800DAZ89VABJPBD93/2023-06-30/ESEF/DK/0/213800DAZ89VABJPBD93-2023-06-30-da/reports/213800DAZ89VABJPBD93-2023-06-30-da.xhtml

So you can get the non-viewer version just by switching the URL.

lukepint-wk:
Closed with the addition of `?disable-viewer` in #579

PR #579, opened 2023-10-23
Add "disable-viewer" parameter.

paulwarren-wk:
#### Reason for change

Requested in #571 .

#### Description of change

A new `disable-viewer` parameter can be used to disable the viewer from loading so that the underlying iXBRL document can be seen directly.  If this parameter is present, and has any value other than `false`, the viewer will stop loading.

In stub viewer mode, this flag will cause the viewer to redirect to the iXBRL document (or the first iXBRL document, if there's more than one).  

#### Steps to Test

Add `?disable-viewer` to the URL for any document.  The viewer should not load, leaving you with the original iXBRL document loaded in the browser.

**review**:
@Arelle/arelle
@paulwarren-wk


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -114,6 +114,11 @@ iXBRLViewer.prototype.isReviewModeEnabled = function () {
     return this.isFeatureEnabled('review');
 }
 
+iXBRLViewer.prototype.isViewerEnabled = function () {
+    const urlParams = new URLSearchParams(window.location.search);
+    return (urlParams.get('disable-viewer') ?? 'false') === 'false'
```
lukepint-wk, line 119: Should this optional functionality be included in the README? 

Similarly, I was wondering if we should add anything to the README for available command line options. Like, `--use-stub-viewer`?

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -193,8 +198,12 @@ iXBRLViewer.prototype._checkDocumentSetBrowserSupport = function () {
 }
 
 iXBRLViewer.prototype.load = function () {
-    var iv = this;
-    var inspector = this.inspector;
+    const iv = this;
+    const inspector = this.inspector;
+
+    if (!this.isViewerEnabled()) {
```
lukepint-wk, line 204: This disables the viewer as mentioned in #571 , however, it was a miscommunication on my part. I considered the "viewer" to be everything around the original document. What I intended to request was to disable everything except the `viewer-pane`. 

The goal would be to still load the original html file with no inspector or top bar.

paulwarren-wk: That should be exactly what this does.  What happens when you try this?

lukepint-wk: It resulted in a blank page when I tried. It looks like it will never enter the `setTimeout` since it does the return when disabled.

paulwarren-wk: That'd explain it.  We should probably make it throw an error or just ignore the option completely if you try to use it in stub viewer mode.

lukepint-wk: That sounds like a good idea and would be less confusing overall IMO.

I switched the gui plugin to not use stub viewer and things worked as expected.

Tangentially, it seems like [this](https://github.com/Arelle/ixbrl-viewer/blob/dc8dcde75239693e83f0ad796eae996556190e2f/iXBRLViewerPlugin/iXBRLViewer.py#L437C19-L437C19) should be updated to `ixbrlviewer.html` instead? Otherwise, the plugin opens to a 404 not found due to the missing `i`.

paulwarren-wk: Thinking some more, I think we can just make this work in stub viewer mode, as we can just make it redirect to the iXBRL document.  Done in a492818

> Tangentially, it seems like [this](https://github.com/Arelle/ixbrl-viewer/blob/dc8dcde75239693e83f0ad796eae996556190e2f/iXBRLViewerPlugin/iXBRLViewer.py#L437C19-L437C19) should be updated to ixbrlviewer.html instead? Otherwise, the plugin opens to a 404 not found due to the missing i.

Yes, I think that would make sense, but I think we should tackle it separately as I'm not sure what else might be relying on that naming convention.

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
paulwarren-wk: Docs for this added in [a492818](https://github.com/Arelle/ixbrl-viewer/pull/579/commits/a49281822746ff8899906e140fa46ba360d9c554)

README improvements are in #580

austinmatherne-wk (approved):

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
lukepint-wk: 👍 , might just want to update the PR description that stub viewer is supported now for posterity.

lukepint-wk (approved):

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
paulwarren-wk: Good call - done.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #589, opened 2023-11-17
Update legacy/VUID fact ID usage for SHOW_FACT and SEC/ESEF hidden facts

aaroncameron-wk:
#### Reason for change
- The `SHOW_FACT` event was passing legacy fact IDs to `selectItem`
- Handling of SEC/ESEF links-to-hidden IDs was not using VUIDs

#### Description of change
Fix above cases where VUIDs were not being used, and do widespread renaming of `id` => `vuid` for clarity of where VUIDs are expected.

#### Steps to Test
CI

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/inspector.js
+++ b/iXBRLViewerPlugin/viewer/src/js/inspector.js
@@ -131,7 +131,7 @@ export class Inspector {
     }
 
     initializeViewer() {
-        this._viewer.onSelect.add((id, eltSet, byClick) => this.selectItem(id, eltSet, byClick));
+        this._viewer.onSelect.add((vuid, eltSet, byClick) => this.selectItem(vuid, eltSet, byClick));
         this._viewer.onMouseEnter.add((id) => this.viewerMouseEnter(id));
         this._viewer.onMouseLeave.add(id => this.viewerMouseLeave(id));
         $('.ixbrl-next-tag').click(() => this._viewer.selectNextTag(this._currentItem));
@@ -889,7 +889,7 @@ export class Inspector {
 
         // dissolveSingle => title not shown if only one item in accordian
         const a = new Accordian({
-            onSelect: (id) => this.switchItem(id),
+            onSelect: (vuid) => this.switchItem(vuid),
             alwaysOpen: true,
             dissolveSingle: true,
         });
@@ -1042,9 +1042,9 @@ export class Inspector {
      * If itemIdList is omitted, the currently selected item list is reset to just
      * the primary item.
      */
-    selectItem(id, itemIdList, noScroll) {
+    selectItem(vuid, itemIdList, noScroll) {
         if (itemIdList === undefined) {
-            this._currentItemList = [ this._reportSet.getItemById(id) ];
+            this._currentItemList = [ this._reportSet.getItemById(vuid) ];
         }
         else {
             this._currentItemList = [];
@@ -1052,7 +1052,7 @@ export class Inspector {
                 this._currentItemList.push(this._reportSet.getItemById(itemId));
             }
         }
-        this.switchItem(id, noScroll);
+        this.switchItem(vuid, noScroll);
     }
 
     /*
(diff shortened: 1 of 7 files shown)


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -173,9 +171,14 @@ export class Inspector {
             // messages to itself when exporting files.
             return;
         }
-
-        if (data.task == 'SHOW_FACT') {
-            this.selectItem(data.factId);
+        if (String(data.task) === SHOW_FACT) {
+            const factId = String(data.factId);
```
brettkail-wk, line 175: Why `String()`?

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -173,9 +171,14 @@ export class Inspector {
...
+            const vuid = this.parseVuid(factId, true);
```
brettkail-wk, line 176: It seems like we should add `data.docSetId` instead of speculatively parsing the data factId.

@joyreistad-wf Can you open a ticket for us to make a change to our internal service using API?

joyreistad-wf: Sure!

aaroncameron-wk: The goal with this is to assume that if we're receiving a fact ID in the legacy format (`f-id1`) that the API user doesn't expect multi-doc support, so defaulting to a report index of 0 should be fine. 

The scenario of `{"docSetId": 0, "factId": "f-id1"}`, doesn't make much sense because `factId` is still in a format we don't intend to continue supporting. 

I'd think we do either `{"docSetId": 0, "factId": "id1"}` and have ixbrl-viewer generate the VUID based off of that, or just do `{"factId": "f0-id1"}`. I'm fine either way, but the latter does not require another API update.

brettkail-wk: AFAIK, SHOW_FACT expects plain XML IDs (`"id1"`) rather than ixbrl-viewer-qualified (`"f-id1"`); at least, I'm not aware of code in our internal service that uses SHOW_FACT that would be prepending that `f-`.  It seems like the simplest bug fix is to add just add `viewerUniqueId(0, str)`, and then we can separately figure out the contract for multi-IXDS.

For multi-IXDS, I prefer the `{"docSetId":0,"factId":id1"}` rather than needing to couple to the VUID format in callers, but I guess I don't feel too strongly.

aaroncameron-wk: Ah, I see. That was a misunderstanding on my part that the `f-id1` format was being used. In that case I agree with the `docSetId` approach.

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -180,3 +184,14 @@ export function viewerUniqueId(sourceReportIndex, localId) {
 export function localId(viewerUniqueId) {
     return viewerUniqueId.replace(/^\d+-/,"");
 }
+
+export function getIXHiddenLinkStyle(domNode) {
```
paulwarren-wk, line 188:
```suggestion
/**
 * Parses fact IDs from -sec-ix-hidden and -esef-ix-hidden style 
 * attributes on a DOM node.  
 * Any returned ID should be the ID of a fact in ix:hidden corresponding
 * to the content contained in the DOM node.
 * 
 * @param  {Node}     domNode - DOM node to parse
 * @return {String}   A fact ID, or null if the element does not have 
 * a style attribute containing a custom CSS property in the required 
 * format.
 */

export function getIXHiddenLinkStyle(domNode) {
```

paulwarren-wk (approved):

derekgengenbacher-wf (approved):

austinmatherne-wk (approved):

derekgengenbacher-wf:

derekgengenbacher-wf:

Issue #576, opened 2023-10-19
[REQUEST] For UTR units, display the unit's name rather than unit's qname in Fact properties


strr:
### What should we change and why?

At the moment, for any unit other than a currency, it is displayed in QName form.

Please consider using the metadata from [http://www.xbrl.org/utr/utr.xml](the UTR) so that instead of displaying a fact value as 
`Fact Value: 15,000 utr:MWh`

They are instead formatted as something like:
`Fact Value: 15,000 Megawatt-Hour` (i.e. lookup the `unitId` in the UTR and output the `unitName` in its place)

paulwarren-wk:
Would it not be more appropriate to use the unit symbol here, rather than the name?

Perhaps we could put the name in a mouseover.

The UTR would also give us symbols for all the other currencies, but I think there are some subtleties around whether the symbol is normally before or after the value.

strr:
Symbol rather than name is fine with me.

aaroncameron-wk:
@paulwarren-wk 

These are some work-in-progress branches I just rebased from a while back. They make a number of related changes, including rendering a symbol when available. It also includes a tooltip for units on hover, and opens a unit legend window on click (although the rendering there is incomplete). I likely won't be able to prioritize work on this so feel free to use directly or simply as a suggestion.

https://github.com/aaroncameron-wk/arelle-public/tree/utr-entries
https://github.com/aaroncameron-wk/ixbrl-viewer/tree/utr-entries

paulwarren-wk:
> These are some work-in-progress branches I just rebased from a while back. 

@aaroncameron-wk - thanks that's really helpful!  It solves I hit with getting an incomplete UTR map for complex entries.

I'll try to merge your UI changes with mine, as I was just using title attributes for the tooltip.

PR #592, opened 2023-11-29
Display complex units correctly

### LINKED ISSUE #575: [BUG] The denominator of a complex unit is not displayed in the fact properties
### What happened?

The denominator of a complex unit is not displayed in the fact properties.

Attached is a sample package
[xii-sig-dsd-example-22 (5).zip](https://github.com/Arelle/ixbrl-viewer/files/13045068/xii-sig-dsd-example-22.5.zip)

It works fine in Arelle's own GUI and you can see both units separated by a '/' in the Fact list
Arelle-GUI-Fact-List-screenshot

In Fact properties in the ixbrl-viewer, you just get the numerator unit only:

Arelle-ixbrl-viewer-unit-missing-denominator

### Version

1.4.3-6-g7f768ea

Microsoft Edge

### Documents

_No response_

### Screenshots

_No response_

aaroncameron-wk:
#### Reason for change
Resolves #575

#### Description of change
Replaces the use of `measureLabel` (which simply took the first numerator of a complex unit for displaying a label) with `unitLabel`, which transforms simple or complex units into title-cased, more human-readable labels with currency symbols where applicable.

#### Steps to Test
Use example doc in linked issue to confirm complex units show completely and correctly:
Screenshot 2023-11-29 at 12 29 14 PM
Screenshot 2023-11-29 at 12 29 08 PM

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/fact.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.js
@@ -109,14 +109,6 @@ export class Fact {
     }
 
     /**
-     * Returns the qname of the first numerator in the fact's unit
-     * @return {String} QName string of a measure
-     */
-    measure() {
-        return this.unit()?.measure();
-    }
-
-    /**
      * Returns details about this fact's unit
      * @return {Unit} Unit instance
      */
--- a/iXBRLViewerPlugin/viewer/src/js/fact.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.js
@@ -77,7 +77,7 @@ export class Fact {
             else {
                 formattedNumber = formatNumber(v, d);
             }
-            if (this.isMonetaryValue() && this.unit().isSimple()) {
+            if (this.isMonetaryValue()) {
                 v = this.unitLabel() + " " + formattedNumber;
             }
             else {
--- a/iXBRLViewerPlugin/viewer/src/js/fact.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.js
@@ -78,10 +78,10 @@ export class Fact {
                 formattedNumber = formatNumber(v, d);
             }
             if (this.isMonetaryValue()) {
-                v = this.measureLabel() + " " + formattedNumber;
+                v = this.unitLabel() + " " + formattedNumber;
             }
             else {
-                v = formattedNumber + " " + this.measureLabel();
+                v = formattedNumber + " " + this.unitLabel();
             }
         }
         else if (this.isNil()) {
@@ -117,14 +117,6 @@ export class Fact {
     }
 
     /**
-     * Returns a readable label representing the first numerator in the fact's unit
-     * @return {String} Label representing measure
-     */
-    measureLabel() {
-        return this.unit()?.measureLabel() ?? i18next.t("factDetails.noUnit");
-    }
-
-    /**
      * Returns details about this fact's unit
      * @return {Unit} Unit instance
      */
@@ -137,6 +129,14 @@ export class Fact {
             this._unit = this.report.reportSet.getUnit(unitKey);
         }
         return this._unit;
+    }
+
+    /**
+     * Returns a readable label representing the fact's unit
+     * @return {String} Label representing unit
+     */
+    unitLabel() {
+        return this.unit()?.label() ?? i18next.t("factDetails.noUnit");
     }
 
     getConceptPrefix() {
--- a/iXBRLViewerPlugin/viewer/src/js/fact.js
+++ b/iXBRLViewerPlugin/viewer/src/js/fact.js
@@ -77,7 +77,7 @@ export class Fact {
             else {
                 formattedNumber = formatNumber(v, d);
             }
-            if (this.isMonetaryValue()) {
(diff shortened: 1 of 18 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/unit.js
```diff
@@ -13,13 +12,17 @@ export class Unit {
                 .split('/');
         this._numerators = split[0].split('*');
         this._denominators = split.length > 1 ? split[1].split('*') : [];
-        this._isMonetary = Boolean(this._numerators.find(n => this._report.qname(n).namespace === "http://www.xbrl.org/2003/iso4217"));
-        this._label = split.map(x => {
-                const part = x.split('*').map(y => {
-                    return titleCase(y.split(':')[1]);
-                }).join('*');
+        this._isSimple = this._numerators.length === 1 && this._denominators.length === 0;
+        this._isMonetary = Boolean(this._numerators.find(n => this._report.qname(n).namespace === NAMESPACE_ISO4217));
```
paulwarren-wk, line 16: Pre-existing, but I think a monetary unit should be a simple unit (single numerator, no denominator) with the measure being ISO 4217 namespace.

iXBRLViewerPlugin/viewer/src/js/fact.js
```diff
@@ -77,11 +77,11 @@ export class Fact {
             else {
                 formattedNumber = formatNumber(v, d);
             }
-            if (this.isMonetaryValue()) {
-                v = this.measureLabel() + " " + formattedNumber;
+            if (this.isMonetaryValue() && this.unit().isSimple()) {
```
paulwarren-wk, line 80: See separate comment - I think "simple" should be part of the definition of monetary.

iXBRLViewerPlugin/viewer/src/js/unit.js
```diff
@@ -48,19 +60,6 @@ export class Unit {
         return this._measure;
     }
```
paulwarren-wk, line 61: I was wondering if this could be removed, but it's still used by `Fact.getScaleLabel()`.  Looking at it, it seems that it only gets used if the fact is monetary, so I think if we update `isMonetary()` to require `isSimple()` as per other comments, then this is fine.

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -207,3 +210,20 @@ export function getIXHiddenLinkStyle(domNode) {
...
+ */
+
+export function measureLabel(report, measure) {
+    const qname = report.qname(measure);
+    if (qname.namespace === NAMESPACE_ISO4217) {
+        measure = i18next.t(`currencies:unitFormat${qname.localname}`, {defaultValue: qname.localname});
+    } else if (measure.includes(':')) {
+        measure = titleCase(measure.split(':')[1]);
+    } else {
+        measure = titleCase(measure);
```
paulwarren-wk, line 226: I don't think that we should change the casing of units.  We have some units that are distinguished only by case (e.g. `Gt` vs `GT` for metric vs imperial gigatonnes)

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -207,3 +210,20 @@ export function getIXHiddenLinkStyle(domNode) {
...
+ * Transforms measure qname into title case label (or currency symbol, if applicable).
+ * @return {String} Measure Label
+ */
+
+export function measureLabel(report, measure) {
+    const qname = report.qname(measure);
+    if (qname.namespace === NAMESPACE_ISO4217) {
+        measure = i18next.t(`currencies:unitFormat${qname.localname}`, {defaultValue: qname.localname});
+    } else if (measure.includes(':')) {
+        measure = titleCase(measure.split(':')[1]);
```
paulwarren-wk, line 224: I actually quite like the idea of including the prefixes for custom units so it'd be clear if a preparer has created their own units, but a) this never happens in practice b) we should treat other namespaces as "standard" (`xbrli` and `utr`) so I think on balance it is better to remove all prefixes for now, as per this change.

There's been some discussion about making more use of UTR definitions in the viewer, so we could perhaps do it as part of that.  In theory, UTR validation prevents misuse of custom units in most cases.

aaroncameron-wk: I was going down this line of thinking as well. If we established a mapping of `utr:XYZ` to symbols, we would have the opportunity for improved control of unit display (e.g. `m³` instead of `m3`). This mapping could be static, or it could be dynamically loaded by Arelle, which is already building UTR entries details (with a couple of changes, UTR `definition` could be brought over as well to be used in a tooltip or unit legend).

For any custom units that wouldn't have a match in this mapping, we could render the full `prefix:unit` display for clarity (however rarely custom units are used).

aaroncameron-wk: I just saw #576, so I anticipate this will all be addressed with that solution.

austinmatherne-wk (approved):

PR #599, opened 2023-12-13
Multi target document support

paulwarren-wk:
#### Reason for change

To support an XBRL feature used by a growing number of jurisdictions

#### Description of change

The viewer builder will now include all target documents that have been loaded by Arelle, and the viewer no longer ignores non-default target documents when pre-processing.

The key changes on the builder side are to create a single "source report" for a given set iXBRL document set, and to change where we do the `deepcopy` of the XML document in order to ensure that the addition of missing ID attributes for all targets make it into the output.

This PR also:

* Fixes namespace highlighting if there are more than three namespaces
* Changes the search results count to refer to "facts" rather than "concepts"

#### Steps to Test

Create a viewer for any document with multiple target documents.  For example, this filing:

https://filings.xbrl.org/filing/529900T6WNZXD2R3JW38-2021-09-30-ESEF-DK-0 (base taxonomies [here](https://erhvervsstyrelsen.dk/vejledning-teknisk-vejledning-og-dokumentation-regnskab-20-taksonomier-historiske))

Recent versions of Arelle will load all targets by default.

Check the namespaces loaded into the viewer (note that there's only one non-hidden, DKGAAP concept in the above example).

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -524,9 +524,13 @@ class iXBRLViewer:
         self.filingDocuments = None
         # This is an arbitrary ModelXbrl used for logging only
         self.logger_model = logger_model
+        self.filenames = set()
 
     def addFile(self, ivf):
+        if ivf.filename in self.filenames:
+            return
         self.files.append(ivf)
+        self.filenames.add(ivf.filename)
 
     def addFilingDoc(self, filingDocuments):
         self.filingDocuments = filingDocuments
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -390,7 +390,6 @@ class IXBRLViewerBuilder:
         self.taxonomyData["sourceReports"].append(sourceReport)
         return sourceReport
 
-
     def createViewer(self, scriptUrl: str = DEFAULT_VIEWER_PATH, useStubViewer: bool = False, showValidations: bool = True, packageDownloadURL: str = None) -> Optional[iXBRLViewer]:
         """
         Create an iXBRL file with XBRL data as a JSON blob, and script tags added.
@@ -408,7 +407,6 @@ class IXBRLViewerBuilder:
         self.roleMap.getPrefix(XbrlConst.parentChild, "pres")
         self.roleMap.getPrefix(XbrlConst.dimensionDefault, "d-d")
         self.roleMap.getPrefix(WIDER_NARROWER_ARCROLE, "w-n")
-
 
         sourceReportsByFiles = dict()
 
@@ -482,7 +480,6 @@ class IXBRLViewerBuilder:
                 localDoc: sorted(docTypes)
                 for localDoc, docTypes in localDocs.items()
             }
-
 
         self.taxonomyData["prefixes"] = self.nsmap.prefixmap
         self.taxonomyData["roles"] = self.roleMap.prefixmap
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -413,7 +413,7 @@ class IXBRLViewerBuilder:
         sourceReportsByFiles = dict()
 
         if useStubViewer:
-            iv.addFile(iXBRLViewerFile(DEFAULT_OUTPUT_NAME, self.getStubDocumet()))
+            iv.addFile(iXBRLViewerFile(DEFAULT_OUTPUT_NAME, self.getStubDocument()))
 
         for n, report in enumerate(self.reports):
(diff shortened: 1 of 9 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -411,59 +409,57 @@ def createViewer(self, scriptUrl: str = DEFAULT_VIEWER_PATH, useStubViewer: bool
...
-                    iv.addFile(iXBRLViewerFile(filename, report.modelDocument.xmlDocument))
+                    filename = srcFilename
+                iv.addFile(iXBRLViewerFile(filename, report.modelDocument.xmlDocument))
+
+            docSetKey = frozenset(docSetFiles)
+            sourceReport = sourceReportsByFiles.get(docSetKey)
+            if sourceReport is None:
+                sourceReport = self.addSourceReport()
+                sourceReportsByFiles[docSetKey] = sourceReport
+                sourceReport["docSetFiles"] = list(urllib.parse.quote(f) for f in docSetFiles)
```
brettkail-wk, line 460: Does this change the format of the JSON (previously plain strings, now URL-quoted strings)?  That seems unfortunate.  It also seems generally odd to store pre-quoted strings in the JSON.

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -375,59 +385,66 @@ export class Viewer {
...
+                    if (isFootnote) {
+                        nodes.addClass("ixbrl-element-footnote");
+                        ixn.footnote = true;
                     }
                 }
-                if (isFootnote) {
-                    nodes.addClass("ixbrl-element-footnote");
-                    ixn.footnote = true;
+            }
+            else if (isContinuation) {
```
brettkail-wk, line 429: Out of curiosity, why split `isContinuation` to its own block?  Is there some functional change I'm not seeing, or is it just for clarity?  If the code is the same except for ID handling, perhaps a small helper function would be useful, and perhaps we could split `isFootnote` into its own block too.

iXBRLViewerPlugin/iXBRLViewer.py
paulwarren-wk: This change was introduced here: https://github.com/Arelle/ixbrl-viewer/pull/394#discussion_r1151188258

The rationale is that `docSetFiles` should be a list of URLs rather than files, hence URL quoting here.

iXBRLViewerPlugin/viewer/src/js/viewer.js
paulwarren-wk: The functional change that precipitated this was switching from checking for absence of `target` attribute to presence of `id` attribute.  The condition at the start of the block was already a bit mad and not really what we wanted: if it's the right element type, but doesn't have an ID it shouldn't be considered by the subsequent `else`s, it should be ignored.  And then the content of the block was mostly conditional anyway, so it seemed clearer to split it.

I've done a bit of refactoring in c1fd3f7 to move common code into the existing helper functions.

Footnotes have a bit more in common with facts, because they get included in the document order index, and they have their own ID associated with them.

iXBRLViewerPlugin/iXBRLViewer.py
brettkail-wk: I did not see the pre-existing `urllib` call, but I see it now, thanks.  I also had not remembered that we made that change, and I guess it was backwards incompatible then too...  Anyway, perhaps it would be helpful to update [the doc](https://github.com/Arelle/ixbrl-viewer/blob/master/ARCHITECTURE.md#docsetfiles) to say it's a list of URLs rather than files.

brettkail-wk (dismissed):

austinmatherne-wk (approved):

iXBRLViewerPlugin/iXBRLViewer.py
paulwarren-wk: I think it was more of a bug fix that a backwards incompatible change, as the viewer code was always treating them as URLs, so if you previously tried to use a filename with a URL-significant character such as `?` or `#` in it, it would break.

I've updated the docs, and in doing so, also remove the note about not using `docSetFiles` for single document IXDS, as this PR also removes that historical anomaly.

brettkail-wk: Alternatively, it could have been considered a bug in the JS.  That way, generated viewers are fixed by just swapping in the new JS rather than needing to be regenerated.  Anyway, thanks for the docs updates.

brettkail-wk (dismissed):

brettkail-wk (approved):

austinmatherne-wk (approved):

PR #611, opened 2024-01-09
Highlight scaled and absolutely positioned facts within tables

austinmatherne-wk:
#### Reason for change
Fact highlighting of absolutely positioned content doesn't apply to facts within tables. Additionally, the highlight outline width is set to a pixel size and doesn't adjust for documents that use CSS scale transforms.

#### Description of change
* Use em sizes to define fact highlight outlines for scaled documents.
* Highlight fact content that is absolutely positioned within tables.
* Fix `AttributeError: 'IXBRLViewerBuilderError' object has no attribute 'message'`
* Use a default empty string script path (for the built-in script file) in the GUI instead of an absolute path so users can have multiple versions of Arelle and ixbrl-viewer (frozen build and Python source) on the same machine.
* Explicitly exit generation early if destination is None.
* Replace nonexistent `os.path.filepart` method reference.
* Update Node versions used for CI tests to current active versions (maintenance, LTS, and current.)

#### Steps to Test
* Test internal doc.

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/__init__.py
+++ b/iXBRLViewerPlugin/__init__.py
@@ -258,10 +258,10 @@ class iXBRLViewerLocalViewer(LocalViewer):
             _fileExists = False
             if os.path.exists(os.path.join(_fileDir, _file)):
                 _fileExists = True
-            elif "/" in _file and os.path.exists(os.path.join(_fileDir, os.path.filepart(_file))):
+            elif "/" in _file and os.path.exists(os.path.join(_fileDir, os.path.basename(_file))):
                 # xhtml in a subdirectory for output files may refer to an image file in parent directory
                 _fileExists = True
-                _file = os.path.filepart(_file)
+                _file = os.path.basename(_file)
             if not _fileExists:
                 self.cntlr.addToLog("http://localhost:{}/{}".format(self.port, file), messageCode="localViewer:fileNotFound", level=logging.DEBUG)
             return static_file(_file, root=_fileDir, headers=self.noCacheHeaders)  # extra_headers modification to py-bottle
--- a/iXBRLViewerPlugin/__init__.py
+++ b/iXBRLViewerPlugin/__init__.py
@@ -9,7 +9,6 @@ import sys
 import tempfile
 import traceback
 from optparse import OptionGroup, OptionParser
-from typing import Optional, Union
 
 from arelle import Cntlr
 from arelle.LocalViewer import LocalViewer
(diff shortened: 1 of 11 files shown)


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/__init__.py
```diff
@@ -203,10 +205,13 @@ def iXBRLViewerSaveCommand(cntlr):
     dialog = SaveViewerDialog(cntlr)
     dialog.render()
     if dialog.accepted and dialog.filename():
+        scriptUrl = dialog.scriptUrl()
+        if scriptUrl == DEFAULT_JS_FILENAME:
+            scriptUrl = DEFAULT_VIEWER_PATH
```
paulwarren-wk, line 210: This might be confusing if a user actually wants a relative URL to `ixbrlviewer.js`.  Would it be better to have a separate "use default" checkbox, or use empty string to get the default with a "leave blank for default" message.

iXBRLViewerPlugin/__init__.py
```diff
@@ -115,6 +115,8 @@ def generateViewer(
     :param features: List of feature names to enable via generated JSON data.
     """
     # extend XBRL-loaded run processing for this option
+    if saveViewerDest is None:
+        return
```
paulwarren-wk, line 119: This is fine, but it leaves me wondering what the later `if out:` is doing.  From the history, we previously had:

```
out = getattr(options, 'saveViewerFile') or kwargs.get("responseZipStream")
```

But that got removed, so I suspect the `if out` is now redundant.

iXBRLViewerPlugin/viewer/src/less/common.less
```diff
@@ -2,12 +2,12 @@
 
 /* Styles common to both inspector and highlighting used within iframe */
 .linked-highlight-text {
-  outline: dashed 2px @linked-fact;
-  outline-offset: 1px;
+  outline: dashed 0.125em @linked-fact;
```
paulwarren-wk, line 5: Nice.  I wish I'd thought of this when I previously tried and failed to fix the issue with massively scaled documents.

iXBRLViewerPlugin/__init__.py
austinmatherne-wk: Updated

iXBRLViewerPlugin/__init__.py
austinmatherne-wk: I wasn't expecting users to enter relative paths since it isn't clear from the GUI where a relative path is starting from, but I went ahead and made it a blank string.

derekgengenbacher-wf (approved):

brettkail-wk (approved):

paulwarren-wk (approved):

derekgengenbacher-wf:

PR #618, opened 2024-01-17
Log error and abort generation if viewer script does not exist

### LINKED ISSUE #617: [BUG] Invalid Viewer URL Doesn't Raise an Error
### What happened?

If you provide a local filesystem viewer url that doesn't exist an error is not raised at generation time and a user only notices when they open the viewer and it fails to load. This is true for both the GUI and command line.

We should log an error if a locale filesystem viewer URL is used and the file does not exist.

### Version

1.4.14

Firefox, Chrome, Safari, Microsoft Edge

### Documents

_No response_

### Screenshots

_No response_

austinmatherne-wk:
#### Reason for change
Resolves #617 

#### Description of change
* Download and copy script into output directory if it's an http URL.
* Log an error and abort generation if script is to a file that doesn't exist or can't be downloaded.
* Add new `--viewer-no-copy-script` (also available in the GUI) option to skip copying or validating `--viewer-url` and trust the user to provide a valid script location.
* Don't open local viewer if generation failed.
* Include standard message codes on all Arelle logs.
* Fix a variable typo.

#### Steps to Test
* Test valid and invalid script URLs from both the GUI and CLI with and without the copy script option.

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/__init__.py
+++ b/iXBRLViewerPlugin/__init__.py
@@ -14,10 +14,11 @@ from pathlib import Path
 from arelle import Cntlr
 from arelle.LocalViewer import LocalViewer
 from arelle.ModelDocument import Type
+from arelle.UrlUtil import isHttpUrl
 from arelle.webserver.bottle import static_file
 
-from .constants import CONFIG_FEATURE_PREFIX, CONFIG_LAUNCH_ON_LOAD, \
-    CONFIG_SCRIPT_URL, DEFAULT_LAUNCH_ON_LOAD, DEFAULT_OUTPUT_NAME, \
+from .constants import CONFIG_COPY_SCRIPT, CONFIG_FEATURE_PREFIX, CONFIG_LAUNCH_ON_LOAD, \
+    CONFIG_SCRIPT_URL, DEFAULT_COPY_SCRIPT, DEFAULT_LAUNCH_ON_LOAD, DEFAULT_OUTPUT_NAME, \
     DEFAULT_JS_FILENAME, DEFAULT_VIEWER_PATH, ERROR_MESSAGE_CODE, \
     EXCEPTION_MESSAGE_CODE, FEATURE_CONFIGS
 from .iXBRLViewer import IXBRLViewerBuilder, IXBRLViewerBuilderError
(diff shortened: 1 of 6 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/__init__.py
```diff
@@ -117,14 +119,22 @@ def generateViewer(
...
         or any(not mx.modelDocument for mx in cntlr.modelManager.loadedModelXbrls)):
-        cntlr.addToLog("No taxonomy loaded.")
+        cntlr.addToLog("No taxonomy loaded.", mesaageCode=ERROR_MESSAGE_CODE)
         return
     modelXbrl = cntlr.modelManager.modelXbrl
     if modelXbrl.modelDocument.type not in (Type.INLINEXBRL, Type.INLINEXBRLDOCUMENTSET):
-        cntlr.addToLog("No inline XBRL document loaded.")
+        cntlr.addToLog("No inline XBRL document loaded.", mesaageCode=ERROR_MESSAGE_CODE)
+        return
+    if not (isHttpUrl(viewerURL) or os.path.exists(viewerURL)):
```
paulwarren-wk, line 131: Won't this break with a relative URL?  A relative URL here would be resolved relative to the CWD, but what matters is whether it exists relative to the output directory.  

Further you might want to use an absolute path here, e.g. `/js/ixbrlviewer.js` which also won't resolve here, but may do once it's behind a web server and accessed via HTTP.

Perhaps we should make this a GUI-only error, and give CLI users more freedom to do stuff that may not work?

I'm also looking at the existing code from line 140, and am a bit confused.  If you give it a relative `viewerURL` it attempts to resolve it relative to the destination directory, and if it exists, it copies it from there.  In the simple case where the relative URL is in the same directory, you end up either copying it onto itself (if it's called `ixbrlviewer.js`) or duplicating it (if it's called anything else).

Maybe what's actually need here are separate, mutually exclusive `viewerURL` and `viewerPath` options: if you specify `viewerPath`, we attempt to copy the script, and reference it as `ixbrlviewer.js` (relative URL), if you specify `viewerURL` we don't attempt to copy the script or modify the URL, and it's the user's problem to make sure that the script is where it should be.

FWIW #614 adds an option to disable the copy script functionality, but I think I'd now prefer the above solution of separate parameters.

austinmatherne-wk: This was a misunderstanding on my part. I wasn't aware that paths relative to the output directory were supported (the puppeteer tests failed because of this).

I investigated adding a `viewerPath` option, but that ran into the hurdle of how to make it optional while also providing the install directory script as a default value so users don't have to figure that out on their own. Ultimately I added a `--viewer-no-copy-script` option (also available in the GUI) to disable copying `--viewer-url` and validating the value. With the new configuration `--viewer-url` is confirmed to exist and downloaded/copied into the output directory (with the basename of the file) unless the `--viewer-no-copy-script` option is used. The negative argument name is unfortunate (I'm open to alternative names), but I think it provides the most intuitive behavior without needing to support 3 arguments. Let me know your thoughts.

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #627, opened 2024-02-11
Display the target document of facts in inspector and search

paulwarren-wk:
#### Reason for change

The viewer will load facts from multiple target documents, but does not show which facts come from with target document.

#### Description of change

Facts from a non-default target document are now tagged as such in search results and the fact inspector:

image

Search:

image

Target document information was not previously included in the report data, so new JS running on old reports will not show target document information.

#### Steps to Test

Load a document with at least one non-default target document, and check that the target document is indicated in search results and the fact inspector.

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/report.js
+++ b/iXBRLViewerPlugin/viewer/src/js/report.js
@@ -39,7 +39,7 @@ export class XBRLReport {
     }
 
     targetDocument() {
-        return this._reportData.target ?? undefined;
+        return this._reportData.target ?? null;
     }
 
     getChildRelationships(conceptName, arcrole) {
--- a/iXBRLViewerPlugin/viewer/src/js/report.js
+++ b/iXBRLViewerPlugin/viewer/src/js/report.js
@@ -38,6 +38,10 @@ export class XBRLReport {
         return this.reportSet.factsForReport(this);
     }
 
+    targetDocument() {
+        return this._reportData.target ?? undefined;
+    }
+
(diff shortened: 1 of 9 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/report.js
```diff
@@ -38,6 +38,10 @@ export class XBRLReport {
         return this.reportSet.factsForReport(this);
     }
 
+    targetDocument() {
+        return this._reportData.target ?? undefined;
```
brettkail-wk, line 42: Out of curiosity, why `?? undefined`?  Is that normalizing `null`->`undefined` and also `undefined` for older serialized viewers?  Or is there a broader convention of using `undefined` for "explicitly assigned to nothing" in JavaScript or this repo?

brettkail-wk (dismissed):

iXBRLViewerPlugin/viewer/src/js/report.js
pdw-mb: Yes, it's to normalize the JSON `null` from new document and absence of property in older viewers.

I don't think we're very consistent on `null` vs `undefined`, but I agree that `null` makes more sense here (now changed).

brettkail-wk (approved):

austinmatherne-wk (approved):

PR #636, opened 2024-02-25
Target document search filter

paulwarren-wk:
#### Reason for change

Allows users to limit searches to specific target documents in multi-target documents.

#### Description of change

Adds a "target documents" filter to the search filters:  

image

This is only shown if there is at least one non-default target document used in the report.

#### Steps to Test

Open a report with a single, default document.  Click on search, then the filters icon, confirm that "Target Documents" filter is not shown.

Open a report with multiple target documents (e.g. Danish ESEF filings).  Confirm that "Target Documents" filter is now shown, and that the relevant facts are shown in the results when selected.

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/inspector.js
+++ b/iXBRLViewerPlugin/viewer/src/js/inspector.js
@@ -384,11 +384,17 @@ export class Inspector {
                 .text(`${prefix} (${this._reportSet.prefixMap()[prefix]})`)
                 .appendTo('#search-filter-namespaces select');
         }
-        for (const targetDocument of this._reportSet.getTargetDocuments()) {
-            $("<option>")
-                .attr("value", targetDocument ?? ':default')
-                .text(targetDocument ?? `<${i18next.t("search.default")}>`)
-                .appendTo('#search-filter-target-document select');
+        const targetDocuments = Array.from(this._reportSet.getTargetDocuments());
+        if (targetDocuments.length == 1 && targetDocuments[0] == null) {
+            $('#search-filter-target-document').hide();
+        }
+        else {
+            for (const targetDocument of targetDocuments) {
+                $("<option>")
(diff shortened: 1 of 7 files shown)


aviary3-wk:
## Security Insights
<details>
<li>Watched keyword <code>style=</code> in <code>iXBRLViewerPlugin/viewer/src/html/inspector.html</code> line(s) ['124'] added</li>
</details>

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/inspector.js
```diff
@@ -384,11 +384,17 @@ export class Inspector {
...
                 .appendTo('#search-filter-namespaces select');
         }
-        for (const targetDocument of this._reportSet.getTargetDocuments()) {
-            $("<option>")
-                .attr("value", targetDocument ?? ':default')
-                .text(targetDocument ?? `<${i18next.t("search.default")}>`)
-                .appendTo('#search-filter-target-document select');
+        const targetDocuments = Array.from(this._reportSet.getTargetDocuments());
+        if (targetDocuments.length == 1 && targetDocuments[0] == null) {
+            $('#search-filter-target-document').hide();
```
austinmatherne-wk, line 389: Is it worth displaying the filter in the case where there's only a single target and it's not the default?

paulwarren-wk: I've remembered my thinking here.  We don't generally hide filters just because there's only a single option, and it's actually useful to show the filter because it makes it explicit that there is only one.

But, many environments don't support multi-target documents at all, and in this case users may well not even know what a target document is and this filter is just confusing clutter, so I think there's value in hiding it (at the cost of it no longer being explicitly clear that there is only one target).

austinmatherne-wk (approved):

PR #640, opened 2024-03-09
Switch to using Intl.DisplayNames

### LINKED ISSUE #573: [REQUEST] Switch to using Intl.DisplayNames class for language names

Spun out of https://github.com/Arelle/ixbrl-viewer/pull/568#discussion_r1347239151

We currently embed names for languages used in the report in the JSON data block, and do not support i18n for these names.

Modern browsers support the [Intl.DisplayNames](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames) class which will give us language names with i18n support.

paulwarren-wk:
Use built-in JS support for multi-lingual labels for languages, rather than including them in the viewer JSON.

Fixes #573

#### Reason for change

We currently include names for languages as part of the viewer JSON and only include them in a single language.  This means that they're not translated as part of the i18n support.

Modern JS provides `Intl.DisplayNames` which can give us language labels in different languages.

#### Description of change

Get labels for languages using `Intl.DisplayNames`.

No longer include `languages` in the taxonomy data object.

New viewers will ignore the `languages` object if present.

#### Steps to Test

Open a viewer, check the list of languages shown on the settings menu.  They should be displayed in the language  of the current locale, but this can be overridden by specifying `?lang=XX` as a query param, where `XX` is a language code.  Where this is done, the language names should be translated.

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -14,7 +14,6 @@ from collections import defaultdict
 from copy import deepcopy
 from pathlib import Path
 
-import pycountry
 from arelle import XbrlConst
 from arelle.ModelDocument import Type
 from arelle.ModelRelationshipSet import ModelRelationshipSet
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -92,7 +92,6 @@ class IXBRLViewerBuilder:
         self.taxonomyData = {
             "sourceReports": [],
             "features": [],
-            "languages": {},
         }
         self.basenameSuffix = basenameSuffix
         self.currentTargetReport = None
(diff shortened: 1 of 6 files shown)


aviary-wf:
## Security Insights

## Action Items
---

brettkail-wk (dismissed):

austinmatherne-wk (dismissed):

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -132,23 +131,6 @@ def escapeJSONForScriptTag(self, s):
         """
         return s.replace("<","\\u003C").replace(">","\\u003E").replace("&","\\u0026")
 
-    def makeLanguageName(self, langCode):
-        code = re.sub("-.*","",langCode)
-        try:
-            language = pycountry.languages.lookup(code)
```
austinmatherne-wk, line 138: `pycountry` is no longer used. Can you drop the import and remove the dependency from `pyproject.toml`?

austinmatherne-wk (approved):

PR #641, opened 2024-03-10
Avoid interfering with layout of tables with absolutely positioned content

### LINKED ISSUE #632: [BUG] Table Style : Position relative is added automatically
### What happened?

The issue is for EU ESEF packages

In the ixbrl viewer, we found that position relative table style is automatically added by the viewer.

This is added to the pages which has text block tags with td tr tags in it.

Please find the attached image for your analysis.

Could you please share your feedback on this please.
image

### Version

2.16.1

Firefox, Chrome, Safari, Microsoft Edge

### Documents

_No response_

### Screenshots

image

paulwarren-wk:
Fixes #632

#### Reason for change

The viewer currently interferes with the layout of tables that contain absolutely positioned content.  The problematic situation is this:

```html
    <div class="page" style="position: absolute">
        <table>
            <tr>
                <td>
                    <div style="position: absolute">1,000</div>
                </td>
            </tr>
        </table>
    </div>
```

The viewer then adds `position: relative` to the `<table>` tag, which means that the absolutely position `<div>` is now positioned with respect to the wrong parent.

The viewer adds this class, because it is used to place the table export icon in the top left of the table.

#### Description of change

The viewer no longer adds an export handle if absolutely positioned content is found within the table.

It would obviously better to enable the export functionality, as the presence of the table tags means that the export should work (where previously it didn't on most ESEF filings), but it's not at all trivial, as the `<table>` tag is likely to be rendered is some random location.

#### Steps to Test

Check that the layout of the sample document on #632 is now correct when loaded in the viewer.

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/tableExport.js
+++ b/iXBRLViewerPlugin/viewer/src/js/tableExport.js
@@ -16,11 +16,26 @@ export class TableExport {
         $('table', iframe).each(function () {
             const table = $(this);
             if (table.find(".ixbrl-element").length > 0) {
-                table.css("position", "relative");
-                const exporter = new TableExport(table, reportSet);
-                $('<div class="ixbrl-table-handle"><span>Export table</span></div>')
-                    .appendTo(table)
-                    .click(() => exporter.exportTable());
+                const containsAbsolute = Array.from(this.querySelectorAll("*"))
+                    .reduce(
+                        (res, elt) => (res || getComputedStyle(elt).getPropertyValue('position') === 'absolute'),
+                        false
+                    );
+
+                // Don't add handles if the table contains absolutely
+                // positioned elements, as doing so will probably interfere
+                // with layout.
+                //
+                // Where a table has absolutely positioned content, it's quite
+                // likely that the table tag itself is positioned in a random
+                // place.
+                if (!containsAbsolute) {
+                    table.css("position", "relative");
+                    const exporter = new TableExport(table, reportSet);
+                    $('<div class="ixbrl-table-handle"><span>Export table</span></div>')
+                        .appendTo(table)
+                        .click(() => exporter.exportTable());
+                }
             }
         });
     }
--- a/samples/src/testdoc/testdoc.ixtmpl
+++ b/samples/src/testdoc/testdoc.ixtmpl
@@ -164,13 +164,18 @@ Profit Loss double tagged: {{ monetary eg:ProfitLoss[] }}{{ monetary eg:ProfitLo
         {{ period cur3mo "2018-09-01" "2018-12-31" }}
         {{ period prev3mo "2017-09-01" "2017-12-31" }}
 
+        <p>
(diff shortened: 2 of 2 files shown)


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/tableExport.js
```diff
@@ -16,11 +16,26 @@ export class TableExport {
         $('table', iframe).each(function () {
             const table = $(this);
             if (table.find(".ixbrl-element").length > 0) {
-                table.css("position", "relative");
-                const exporter = new TableExport(table, reportSet);
-                $('<div class="ixbrl-table-handle"><span>Export table</span></div>')
-                    .appendTo(table)
-                    .click(() => exporter.exportTable());
+                const containsAbsolute = Array.from(this.querySelectorAll("*"))
+                    .reduce(
```
austinmatherne-wk, line 20: Is there a reason to use this implementation of `reduce()` instead of `some()`?

paulwarren-wk: No good one :)
Fixed.

austinmatherne-wk (approved):

iXBRLViewerPlugin/viewer/src/js/tableExport.js
```diff
@@ -16,11 +16,23 @@ export class TableExport {
...
+                const containsAbsolute = Array.from(this.querySelectorAll("*"))
+                    .some((elt) => getComputedStyle(elt).getPropertyValue('position') === 'absolute');
+
+                // Don't add handles if the table contains absolutely
+                // positioned elements, as doing so will probably interfere
+                // with layout.
+                //
+                // Where a table has absolutely positioned content, it's quite
+                // likely that the table tag itself is positioned in a random
+                // place.
```
brettkail-wk, line 28: I worry this will break/remove the handle from Workiva-generated documents, which use `position:absolute` on all `<div>` within a `<table>` but I believe position the `<table>` in a specific/non-random place ([test](https://github.com/Arelle/ixbrl-viewer/files/14559510/p-2024-03-11-en.zip)).

brettkail-wk: I saw a link to [this blog post](https://frontendmasters.com/blog/what-you-need-to-know-about-modern-css-spring-2024-edition/#anchor-positioning) over the weekend, which seemed promising:
> Once you can use this freely, you’ll have to care less about exact DOM positioning of elements (aside from accessibility concerns). The way it is now, the element you want to position relative to another has to be a child element and for there to be a positioning context to work within. This can dictate where elements go in the DOM, whether or not that makes sense.

Looks like Firefox [intends to prototype](https://bugzilla.mozilla.org/show_bug.cgi?id=1838746), which is promising.

PR #680, opened 2024-05-22
Display extracted software vendor names in summary panel

aaroncameron-wk:
#### Reason for change
To allow for more prominent display of software credits associated with a filing that would otherwise be only visible in comments.

#### Description of change
- For each report, retrieve software vendor names from each associated document and store in `softwareCredits` within corresponding JSON object
- When rendering a report summary, display each software vendor name under "Report Creation" section

#### Steps to Test
Example filing with single name match: 
[credittest_name_match.zip](https://github.com/Arelle/ixbrl-viewer/files/15410350/credittest_name_match.zip)
Screenshot 2024-05-22 at 7 17 12 PM

Example filing with multiple pattern matches:
[credittest_pattern_match.zip](https://github.com/Arelle/ixbrl-viewer/files/15410354/credittest_pattern_match.zip)
Screenshot 2024-05-22 at 7 16 55 PM

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -407,11 +407,12 @@ class IXBRLViewerBuilder:
         for n, report in enumerate(self.reports):
             self.footnoteRelationshipSet = ModelRelationshipSet(report, "XBRL-footnotes")
             self.currentTargetReport = self.newTargetReport(getattr(report, "ixdsTarget", None))
-            softwareCredits = {
-                match
-                for document in report.urlDocs.values()
-                for match in document.creationSoftwareMatches(document.creationSoftwareComment)
-            }
+            softwareCredits = set()
+            for document in report.urlDocs.values():
+                if document.type not in (Type.INLINEXBRL, Type.INLINEXBRLDOCUMENTSET):
+                    continue
+                matches = document.creationSoftwareMatches(document.creationSoftwareComment)
+                softwareCredits.update(matches)
             if softwareCredits:
                 self.currentTargetReport["softwareCredits"] = list(softwareCredits)
             for f in report.facts:
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -407,6 +407,13 @@ class IXBRLViewerBuilder:
         for n, report in enumerate(self.reports):
             self.footnoteRelationshipSet = ModelRelationshipSet(report, "XBRL-footnotes")
             self.currentTargetReport = self.newTargetReport(getattr(report, "ixdsTarget", None))
+            softwareCredits = {
+                match
+                for document in report.urlDocs.values()
+                for match in document.creationSoftwareMatches(document.creationSoftwareComment)
+            }
+            if softwareCredits:
+                self.currentTargetReport["softwareCredits"] = list(softwareCredits)
             for f in report.facts:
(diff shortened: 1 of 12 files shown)


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -407,6 +407,13 @@ def createViewer(
         for n, report in enumerate(self.reports):
             self.footnoteRelationshipSet = ModelRelationshipSet(report, "XBRL-footnotes")
             self.currentTargetReport = self.newTargetReport(getattr(report, "ixdsTarget", None))
+            softwareCredits = {
+                match
+                for document in report.urlDocs.values()
```
austinmatherne-wk, line 412: I think this is going to include creation software used for both the report and any taxonomy files (including filer extension and base taxonomy). We should limit this to just the inline report documents.

iXBRLViewerPlugin/viewer/src/js/report.js
```diff
@@ -266,4 +266,17 @@ export class XBRLReport {
         }
         return this._calculationSummations.has(c);
     }
+
+
+    /**
+     * @return {Array[String]} Software credit text labels provided with this report for display purposes.
+     */
+    softwareCredits() {
+        if (!this._reportData.softwareCredits) {
```
austinmatherne-wk, line 275: Is this going to throw an exception if the JS is used with a viewer that was generated with an older version of the python plugin that didn't include softwareCredits?

iXBRLViewerPlugin/viewer/src/html/inspector.html
```diff
@@ -346,6 +346,12 @@ <h3 class="collapsible-header" data-i18n="inspector.summary.file.header">File Su
               </div>
             </div>
           </div>
+          <div class="collapsible-section software">
+            <h3 class="collapsible-header" data-i18n="inspector.summary.filingSoftware">Filing Software</h3>
```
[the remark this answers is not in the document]
aaroncameron-wk: Updated to "Report Creation". The thought process being that those credits will always be specific to the generation/creation of the report, but may not always actually refer to software (a consultant, perhaps)? Wide open to alternative suggestions.

iXBRLViewerPlugin/viewer/src/js/report.js
```diff
@@ -266,4 +266,17 @@ export class XBRLReport {
...
+            return [];
+        }
+        // Arelle defaults to "None" if no comment text is found. This will be null for our purposes.
+        return this._reportData.softwareCredits
+                .filter(c => c !== "None");
```
brettkail-wk, line 280: It's unfortunate that the model returns `'None'` rather than `None`.  In any case, it seems like this should be filtered out in the Python instead.

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/report.js
aaroncameron-wk: I suppose with the Arelle `creationSoftwareMatches` refactor I did before this ticket, this filter is no longer necessary anywhere (the main point of the refactor was to decouple the software name matching from undesired existing behavior of returning "None" or spitting out the full comment text)

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #684, opened 2024-05-28
Try to use xml:lang to set html lang

austinmatherne-wk:
#### Reason for change
The default lang value of `en-US` potentially conflicts with `xml:lang` if present.

#### Description of change
First check if there's an `xml:lang` value to use before falling back to `en-US` when setting `lang`.

#### Steps to Test
Note: this doesn't impact the stub viewer, so this must be tested from the command line.
* Use [a report with a non English xml:lang value](https://github.com/Arelle/ixbrl-viewer/files/15476117/austin-2024-05-28-fr.zip).
* `python arelleCmdLine.py --plugins ixbrl-viewer --file austin-2024-05-28-fr.zip --save-viewer ixbrlviewer.html`
* Load the viewer in a web browser and confirm that the `xml:lang` and `lang` attributes on the html element match (`fr`).

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
@@ -157,7 +157,7 @@ export class iXBRLViewer {
             docTitle = "Inline Viewer";
         }
         if ($('html').attr("lang") === undefined) {
-            $('html').attr("lang", "en-US");
+            $('html').attr("lang", $('html').attr("xml:lang") || "en-US");
         }
 
         $('head').children().not("script").not("style#ixv-style").not("link#ixv-favicon").appendTo($(iframe).contents().find('head'));


aviary3-wk:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -157,7 +157,7 @@ export class iXBRLViewer {
             docTitle = "Inline Viewer";
         }
         if ($('html').attr("lang") === undefined) {
-            $('html').attr("lang", "en-US");
+            $('html').attr("lang", $('html').attr("xml:lang") || "en-US");
```
brettkail-wk, line 160: It looks like the code originated from [this PR](https://github.com/Arelle/ixbrl-viewer/pull/184), and there was no particular semantic being tied to the attribute (e.g., for translations).

It seems somewhat odd to default to `en-US` at all: if the document doesn't declare a global `xml:lang`, I wonder if it would be better to just do nothing?

brettkail-wk (approved):

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
austinmatherne-wk: Dropping the `en-US` fallback and doing nothing if `xml:lang` isn't defined makes sense to me. @paulwarren-wk do you have any objections?

paulwarren-wk: I assume that this was added because the accessibility report said we should have a `lang` attribute, and at the time `en-US` was the only language supported by the UI – this attribute is on the parent `<html>` not the iframe.

I think we should be explicitly setting `lang` to the result of `selectDefaultLanguage()` for all viewers, not just when reparenting.

When reparenting, we should probably copy `xml:lang` from the parent to the child document, and then delete it from the parent.

austinmatherne-wk: As I understand it, there are three distinct but related languages in an instance of an Arelle viewer:

1. The language of the HTML content.
2. The languages of the XBRL taxonomy.
3. The language of the Arelle viewer UI.

A user opening the viewer can express a language preference in three ways:

1. Selecting a language from the gear dropdown menu.
2. Using a `lang` query parameter.
3. Through the browser's language settings.

The language of the HTML content remains constant for each report and is unaffected by the user's settings. If the document declares a `lang` attribute, we should copy it to `xml:lang`; otherwise, we should not set the HTML content language. When re-parenting, we should transfer the HTML language attributes to the child iframe.

For the taxonomy labels and viewer UI, we should prioritize displaying:

1. The language selected by the user in the gear menu.
2. The language specified in the query parameter.
3. The languages from the browser's settings.
4. An English fallback.

For taxonomy labels, there is an additional fifth level: if the taxonomy doesn't have English labels, we should select the first language code from the sorted list of taxonomy language codes.

Does this approach seem sensible to both of you?

paulwarren-wk: I had previously assumed that the gear menu would only affect the display of the taxonomy labels, and not the viewer UI, although on reflection, I'm not sure that there's much logic to that position.

I think there one unresolved question: should the viewer HTML (i.e. the stub viewer, or the enclosing HTML after the document HTML has been reparented) have a `lang` or `xml:lang` attribute?  I assume that we added this because the accessibility report requested it, so it would be good not to regress on this.  I think we should probably add one, and set it to the language of the viewer UI, but we need to make sure that we don't clobber the document's attribute before copying it into the iframe (in the case of a non-stub document).

austinmatherne-wk: There may be additional complexity in changing the UI language from the gear menu as currently that's only handled during initialization, but it should be possible.

Setting the language attributes on the viewer HTML to reflect the viewer UI language choice is sensible to me.

brettkail-wk: > I had previously assumed that the gear menu would only affect the display of the taxonomy labels, and not the viewer UI, although on reflection, I'm not sure that there's much logic to that position.

I also assumed that position, and I do think there's a logic to it.  I think the in-app setting should only be used to navigate the report, not to change the UI itself.  For example, as an English-only speaker, I might want to select another language just to see the labels appear, but I would not want that to change the localization of the app UI since I don't speak other languages.

> should the viewer HTML (i.e. the stub viewer, or the enclosing HTML after the document HTML has been reparented) have a lang or xml:lang attribute?

Only `xml:lang` is valid iXBRL, and we previously tried to make that true.

PR #686, opened 2024-05-29
Support for Calculation 1.1 relationships

paulwarren-wk:
#### Reason for change

Although we already have a mode for applying Calc 1.1 semantics to legacy calculation relationships, reports using Calculation 1.1 relationships in the taxonomy did not show any calculation relationships in the viewer.

#### Description of change

Calculations 1.1 relationships are now included in the JSON taxonomy data.  The viewer will now show calculations defined using either relationship type, but if a total has both types in a single ELR, Calc 1.1 takes precedence.

#### Steps to Test

Prepare a report using the Calc 1.1 summation-item relationship, and confirm that calculations are shown in the viewer.  Example below is a filing that has been modified to use the the new arcrole:

[wk-calc11.zip](https://github.com/Arelle/ixbrl-viewer/files/15491364/wk-calc11.zip)

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/calculation.js
+++ b/iXBRLViewerPlugin/viewer/src/js/calculation.js
@@ -4,6 +4,7 @@ import Decimal from 'decimal.js';
 import { setDefault } from './util.js';
 import { Interval } from './interval.js';
 import { FactSet } from './factset.js';
+import { CALC11_ARCROLE, CALC_ARCROLE } from './util.js';
 
 export class Calculation {
     
@@ -19,7 +20,7 @@ export class Calculation {
         const report = fact.report;
         if (!this._conceptToFact) {
             const ctf = {};
-            for (const version of ["calc", "calc11"]) {
+            for (const version of [CALC_ARCROLE, CALC11_ARCROLE]) {
                 const rels = report.getChildRelationships(fact.conceptName(), version)
                 for (const [elr, rr] of Object.entries(rels)) {
                     setDefault(ctf, version, {});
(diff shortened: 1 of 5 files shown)


iXBRLViewerPlugin/viewer/src/js/calculation.js
```diff
@@ -18,13 +18,16 @@ export class Calculation {
...
         const report = fact.report;
         if (!this._conceptToFact) {
-            const rels = report.getChildRelationships(fact.conceptName(), "calc")
             const ctf = {};
-            for (const [elr, rr] of Object.entries(rels)) {
-                ctf[elr] = {};
-                if (rr.length > 0) {
-                    const otherFacts = report.getAlignedFacts(fact, {"c": rr.map(r => r.t )});
-                    otherFacts.forEach(otherFact => setDefault(ctf[elr], otherFact.conceptName(), new FactSet()).add(otherFact));
+            for (const version of ["calc", "calc11"]) {
```
brettkail-wk, line 22: Should it be `["calc"]` only if `this.calc11` is false?

brettkail-wk (dismissed):

iXBRLViewerPlugin/viewer/src/js/calculation.js
paulwarren-wk: I was wondering what to do about this as I didn't want the calculations to simply disappear if you disable Calc 1.1 mode.

I've just pushed 65af4b2 which adds an explicit notice in this case:

image

austinmatherne-wk (approved):

PR #114, opened 2020-06-21
Fix fact deep linking

paulwarren-wk:
When selecting fact, the viewer will update the fragment identifier in the URL of the browser.  Copying the URL and opening should open the viewer with that fact visible and selected.

This was broken by an earlier commit which caused the URL to be updated to a blank fact before the viewer had finished loading, resulting in no fact being selected.  This PR fixes that.

PR #704, opened 2024-07-16
EDGAR workflow-specific operational change

hermfischer-wf:
#### Reason for change
This change is a companion to [EdgarRenderer Pull Request #114](https://github.com/Arelle/EdgarRenderer/pull/114)

With the Edgar workflow the iXBRLViewer stub is generated twice (for the situation of private and redacted public viewers), not by the usual plugin invocation but instead invoked from an EdgarRenderer iXBRLViewerPluginInterface module.

EdgarRenderer provides unique fact IDs because they're also needed by EDGAR's ixviewer-plus.

It adds a stubOnly flag to generateViewer, so that when saving (into an io.bytes zip) only the stub is saved.  The source files are not readable because they may either be encrypted or in a POSTED zip source, but not plain filesystem files.  The source files also may not represent the redacted publicly disseminatable files.

#### Description of change
Add stub-only parameter. 

#### Steps to Test

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/iXBRLViewer.py
+++ b/iXBRLViewerPlugin/iXBRLViewer.py
@@ -546,7 +546,7 @@ class iXBRLViewer:
     def addFilingDoc(self, filingDocuments):
         self.filingDocuments = filingDocuments
 
-    def save(self, destination: io.BytesIO | str, zipOutput: bool = False, copyScriptPath: Path | None = None):
+    def save(self, destination: io.BytesIO | str, zipOutput: bool = False, copyScriptPath: Path | None = None, stubFileOnly: bool = False):
         """
         Save the iXBRL viewer.
         :param destination: The target that viewer data/files will be written to (path to file/directory, or a file object itself).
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Security Insights

## Action Items
---

iXBRLViewerPlugin/iXBRLViewer.py
```diff
@@ -546,7 +546,7 @@ def addFile(self, ivf):
     def addFilingDoc(self, filingDocuments):
         self.filingDocuments = filingDocuments
 
-    def save(self, destination: io.BytesIO | str, zipOutput: bool = False, copyScriptPath: Path | None = None):
+    def save(self, destination: io.BytesIO | str, zipOutput: bool = False, copyScriptPath: Path | None = None, stubFileOnly: bool = False):
```
paulwarren-wk, line 549: I think we should probably catch `stubFileOnly` used without `useStubViewer` and throw an error, as currently you won't get an error, you'll just silently get the first file from the document set.

Perhaps we could move the option on to `createViewer` and use it to control the calls to `addFile`? This might be clearer than just only outputting the first file.

PR #764, opened 2024-11-20
Migrate lang attributes to iframe documents and remove default

austinmatherne-wk:
#### Reason for change
Reopening #684. The previous discussion moved into changing the application language. That feature is being handled by another effort.

#### Description of change
We currently set the lang attribute to a default of en-US for all non stub viewers.

#### Steps to Test
* Use [a report with a non English xml:lang value](https://github.com/Arelle/ixbrl-viewer/files/15476117/austin-2024-05-28-fr.zip).
* Verify that the lang attribute on the root HTML node reflects the application UI language and is updated when you change it.
* Verify that for the non-stub viewer, that all attributes of the report document are migrated to the iframe HTML node for the report document.
* Verify that that the lang attribute on the report document iframe HTML nodes reflects the xml:lang attribute from the filing.

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
@@ -202,8 +202,8 @@ export class iXBRLViewer {
         doc.close();
 
         let docTitle = $('title').text();
-        if (docTitle != "") {
-            docTitle = "Inline Viewer - " + docTitle;
+        if (docTitle !== "") {
+            docTitle = `Inline Viewer - ${docTitle}`;
         }
         else {
             docTitle = "Inline Viewer";
@@ -237,7 +237,7 @@ export class iXBRLViewer {
     _getTaxonomyData() {
         for (let i = document.body.children.length - 1; i >= 0; i--) {
             const elt = document.body.children[i];
-            if (elt.tagName.toUpperCase() == 'SCRIPT' && elt.getAttribute("type") == 'application/x.ixbrl-viewer+json') {
+            if (elt.tagName.toUpperCase() === 'SCRIPT' && elt.getAttribute("type") === 'application/x.ixbrl-viewer+json') {
                 return elt.innerHTML;
             }
         }
@@ -245,7 +245,7 @@ export class iXBRLViewer {
     }
 
     _checkDocumentSetBrowserSupport() {
-        if (document.location.protocol == 'file:') {
+        if (document.location.protocol === 'file:') {
             alert("Displaying iXBRL document sets from local files is not supported.  Please view the viewer files using a web server.");
         }
     }
@@ -258,11 +258,11 @@ export class iXBRLViewer {
             else {
                 fetch(this.options.configUrl)
                     .then((resp) => {
-                        if (resp.status == 404) {
+                        if (resp.status === 404) {
                             return Promise.resolve({});
                         }
-                        else if (resp.status != 200) {
-                            return Promise.reject("Fetch failed: " + resp.status);
+                        if (resp.status !== 200) {
+                            return Promise.reject(`Fetch failed: ${resp.status}`);
                         }
                         return resp.json();
                     })
@@ -300,7 +300,7 @@ export class iXBRLViewer {
             // We need to parse JSON first so that we can determine feature enablement before loading begins.
             const taxonomyData = iv._getTaxonomyData();
             const parsedTaxonomyData = taxonomyData && JSON.parse(taxonomyData);
-            let features = parsedTaxonomyData && parsedTaxonomyData["features"];
+            let features = parsedTaxonomyData?.features;
             if (!features) {
                 features = {};
             }
@@ -349,19 +349,19 @@ export class iXBRLViewer {
             const progress = stubViewer ? 'Loading iXBRL Report' : 'Loading iXBRL Viewer';
             iv.setProgress(progress).then(() => {
                 /* Poll for iframe load completing - there doesn't seem to be a reliable event that we can use */
-                let timer = setInterval(function () {
+                const timer = setInterval(() => {
                     let complete = true;
-                    iframes.each(function (n) {
-                        const iframe = this;
+                    iframes.each((n, iframe) => {
                         const iframeDoc = iframe.contentDocument || iframe.contentWindow.document;
-                        if ((iframeDoc.readyState != 'complete' && iframeDoc.readyState != 'interactive') || $(iframe).contents().find("body").children().length == 0) {
+                        if ((iframeDoc.readyState !== 'complete' && iframeDoc.readyState !== 'interactive') || $(iframe).contents().find("body").children().length === 0) {
                             complete = false;
                         }
                     });
                     if (complete) {
                         clearInterval(timer);
 
-                        const viewer = iv.viewer = new Viewer(iv, iframes, reportSet);
+                        const viewer = new Viewer(iv, iframes, reportSet);
+                        iv.viewer = viewer
 
                         viewer.initialize()
                             .then(() => inspector.initialize(reportSet, viewer))
@@ -382,8 +382,8 @@ export class iXBRLViewer {
                                 .on('resizemove', (event) => {
                                     const target = event.target;
                                     const w = 100 * event.rect.width / $(target).parent().width();
-                                    target.style.width = w + '%';
-                                    $('#inspector').css('width', (100 - w) + '%');
+                                    target.style.width = `${w}%`;
+                                    $('#inspector').css('width', `${100 - w}%`);
                                 })
                                 .on('resizeend', (event) =>
                                     $('#ixv').css("pointer-events", "auto")
@@ -421,7 +421,7 @@ export class iXBRLViewer {
              * message up before the ensuing thread-blocking work
              * https://bugs.chromium.org/p/chromium/issues/detail?id=675795 
              */
-            window.requestAnimationFrame(function () {
+            window.requestAnimationFrame(() => {
                 console.log(`%c [Progress] ${msg} `, 'background: #77d1c8; color: black;');
                 $('#ixv .loader .text').text(msg);
                 window.requestAnimationFrame(() => resolve());
--- a/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
@@ -7,7 +7,7 @@ import { Viewer, DocumentTooLargeError } from "./viewer.js";
 import { Inspector } from "./inspector.js";
 import { initializeTheme } from './theme.js';
 import { TaxonomyNamer } from './taxonomynamer.js';
-import {FEATURE_GUIDE_LINK, FEATURE_REVIEW, FEATURE_SUPPORT_LINK, FEATURE_SURVEY_LINK} from "./util";
+import { FEATURE_GUIDE_LINK, FEATURE_REVIEW, FEATURE_SUPPORT_LINK, FEATURE_SURVEY_LINK, moveNonAppAttributes } from "./util";
 
 const featureFalsyValues = new Set([undefined, null, '', 'false', false]);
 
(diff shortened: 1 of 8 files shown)


iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -252,15 +249,31 @@ export class iXBRLViewer {
...
                         if ((iframeDoc.readyState != 'complete' && iframeDoc.readyState != 'interactive') || $(iframe).contents().find("body").children().length == 0) {
                             complete = false;
                         }
                     });
                     if (complete) {
                         clearInterval(timer);
+                        const html = $('html');
+                        const lang = html.attr('lang') || html.attr('xml:lang');
+                        // Remove any report language from the application html node.
+                        html.removeAttr('lang');
```
brettkail-wk, line 263: `const lang` seems unused.

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -252,15 +249,31 @@ export class iXBRLViewer {
...
+                        html.removeAttr('xml:lang');
+                        iframes.each((n, iframe) => {
+                            const iframeDoc = iframe.contentDocument || iframe.contentWindow.document;
+                            const iframeHtml = $(iframeDoc).find('html');
+                            let frameLang = iframeHtml.attr('lang') || iframeHtml.attr('xml:lang');
+                            if (frameLang === undefined) {
+                                frameLang = lang;
+                            }
+                            if (frameLang !== undefined) {
+                                // Try to set the language of the iframe to the language of the report.
```
brettkail-wk, line 273: I believe this explains what we're trying to do?
> For accessibility, the lang attribute should be set, but XHTML does not allow that, so copy the xml:lang attribute into the lang attribute.

The only detail I'm unsure on that would be nice to have in the comment is whether this is actually required for accessibility tools when the document is served as XHTML (vs HTML).  Perhaps the created iframe can never be served as XHTML?  Or perhaps the dynamically created iframe inherits the XHTML-ness of the parent document?

Anyway, it seems like we can simplify to just this?  Otherwise, I don't understand why `.attr('lang')` is necessary (since it's not valid XHTML) or why we would copy into `xml:lang` (what software would care?).
```
const frameLang = iframeHtml.attr('xml:lang');
if (frameLang !== undefined) {
  iframeHtml.attr('lang', frameLang);
}
```

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
austinmatherne-wk: It's used on line 270.

derekgengenbacher-wf (dismissed):

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
paulwarren-wk: > We currently set the lang attribute to a default of en-US and only for non stub viewers. This change removes the default and migrates the lang attribute to report iframes so as not to contradict the UI language.

Is this definitely the logic we want?  I think it's still only non-stub mode that we care about, in which case we should migrate the parent `xml:lang` attribute, if any, to the first (re-parented) iframe.

For stub viewer mode, and additional non-stub iframes, I think we should leave them untouched.

I don't think we need to check for `lang` attributes because they'd make the document invalid if present (no harm in deleting them just to make sure, though).

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
brettkail-wk: Ah!  I see it now, thank you.

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
austinmatherne-wk: Simple enough. Updated.

derekgengenbacher-wf (dismissed):

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -156,8 +156,9 @@ export class iXBRLViewer {
         else {
             docTitle = "Inline Viewer";
         }
-        if ($('html').attr("lang") === undefined) {
-            $('html').attr("lang", "en-US");
+        if ($('html').attr("xml:lang") !== undefined ) {
+            $(iframe).contents().find('html').attr("xml:lang", $('html').attr("xml:lang"));
+            $('html').removeAttr("xml:lang");
```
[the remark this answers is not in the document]
austinmatherne-wk: The `xml:lang` value defined for the report doesn't necessarily match the language of the viewer UI, so keeping it at the root html level seems inappropriate. The report `xml:lang` value is being moved to the iframe HTML element were the report body is also moved.

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -156,8 +156,9 @@ export class iXBRLViewer {
         else {
             docTitle = "Inline Viewer";
         }
-        if ($('html').attr("lang") === undefined) {
-            $('html').attr("lang", "en-US");
+        if ($('html').attr("xml:lang") !== undefined ) {
+            $(iframe).contents().find('html').attr("xml:lang", $('html').attr("xml:lang"));
```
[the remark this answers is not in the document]
austinmatherne-wk: The lang attribute was likely added to pass an automated accessibility checker that ensures a lang value is defined, but it didn’t verify whether the lang attribute accurately reflects the content.

There’s an open support ticket regarding this issue. I don’t have strong preferences for the final implementation, as long as we stop labeling French content as English.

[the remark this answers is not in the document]
brettkail-wk: The `lang="en-US"` seems like the correct default for the application.  It was always correct prior to commit 9cb2512da0c435bf12198e21104ab015da03eba5, but perhaps it's not correct after `i18nInit` runs.  Perhaps `setLanguage` should also be updating `$('html').attr('lang')` to match the selected application language?

If we need to copy attributes into the iframe (I'm not sure if `xml:lang` has any meaning given that we know from the accessibility ticket that `xml:lang` wasn't good enough.. ?), it seems like we should copy all attributes like we do for `body` below.

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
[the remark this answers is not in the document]
brettkail-wk: Ah, so this seems like the real fix: we want to set `lang` on the root of the application for accessibility, but to avoid confusing (since there's seemingly no real impact?), we should remove the leftover `xml:lang` from the reparented content document.

ixbrl-viewer only supports two application languages, so the user should only get `lang="en"` or `lang="es"` when looking at the DOM, so we probably need to explain that the user is looking in the wrong place to see the language of their report.

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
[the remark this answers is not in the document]
austinmatherne-wk: I tested all major desktop screen readers and observed three distinct behaviors:

1.	All languages are treated the same, using the same voice.
2.	Screen readers inspect the content to determine the language, ignoring both lang and xml:lang attributes (though these might be fallback options if the language can’t be inferred from the content).
3.	The lang attribute is used, while the xml:lang attribute is ignored.

In none of the tools did I find the xml:lang attribute being used. Additionally, the presence of content within an iframe or serving the document as XHTML made no difference to the behavior.

Based on these findings, I propose the following approach:

1.	For the non-stub viewer, if the xml:lang attribute is present, migrate it from the root html node to the html element of the report iframe.
2.	For both stub and non-stub viewers, set the lang attribute of the report iframe’s html element to the value of the report’s xml:lang attribute (if present).
3.	Set the lang attribute of the root html node to match the application UI language.
4.	If the report lacks an xml:lang value, set the lang attribute of the report iframe’s html element to an empty string. This effectively “resets” the language and prompts the screen reader to use its default voice.
5.	Within the inspector, override the lang attribute on taxonomy content elements with the selected taxonomy language.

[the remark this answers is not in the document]
brettkail-wk: > 2. Screen readers inspect the content to determine the language, ignoring both lang
> 3. The lang attribute is used

These seem contradictory.  Can you clarify?

> 5. Within the inspector, override the lang attribute on taxonomy content elements with the selected taxonomy language.

I didn't follow this.  Could you give an example?

Otherwise, 1-4 sound good to me, thanks.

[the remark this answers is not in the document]
austinmatherne-wk: > These seem contradictory.  Can you clarify?

They are. The listed behaviors are from different screen readers. A few examples:
1. NVDA - All languages are treated the same
2. macOS VoiceOver -  inspects the content to determine the language
3. JAWS - Uses the lang attribute

> I didn't follow this.  Could you give an example?

Taxonomy labels, references, etc. There are three languages in the viewer and the lang attributes should reflect that.

1. The report content (report iframe)
2. The selected application UI language (inspector panel)
3. The selected taxonomy language (labels )

iXBRLViewerPlugin/viewer/src/js/ixbrlviewer.js
```diff
@@ -349,19 +348,30 @@ export class iXBRLViewer {
...
                             complete = false;
                         }
                     });
                     if (complete) {
                         clearInterval(timer);
 
-                        const viewer = iv.viewer = new Viewer(iv, iframes, reportSet);
+                        iframes.each((n, iframe) => {
+                            const htmlNode = $(iframe).contents().find('html');
+                            if (htmlNode.attr('lang') === undefined) {
```
brettkail-wk, line 364: This is fine, but this check should never return true since `lang` isn't valid iXBRL, right?  Should we just omit the check or add a comment why we're checking it?

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -232,3 +234,17 @@ export function getIXHiddenLinkStyle(domNode) {
     }
     return null;
 }
+
+/**
+ * Moves all attributes from one element to another, excluding data attributes created by the viewer application.
+ */
+export function moveNonAppAttributes(fromElement, toElement) {
+        // Hypen seperated dataset attributes are returned camel case when accessed via jquery .prop().
```
brettkail-wk, line 242: nit: function body is over-indented.

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -232,3 +234,17 @@ export function getIXHiddenLinkStyle(domNode) {
...
+        const hypenSepAppAttrPrefix = `data-${IXBRL_VIEWER_DATASET_PREFIX.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase()}`;
```
brettkail-wk, line 243: I think it's a DOM `dataset` thing rather than a jQuery thing.

I wonder if copying `.attributes` (but excluding `data-`) and then separately copying `.dataset` would be simpler.

(I wonder if "Designing HTML for Inline XBRL" should recommend a `data-` prefix.)

brettkail-wk (dismissed):

brettkail-wk (approved):

derekgengenbacher-wf (approved):

aaroncameron-wk:
QA +10
- Lang attribute on the root HTML node reflects the application UI language and is updated when you change it.
- For the non-stub viewer, all attributes of the report document are migrated to the iframe HTML node for the report document.
- The lang attribute on the report document iframe HTML nodes reflects the xml:lang attribute from the filing.

PR #810, opened 2025-01-30
Persist App Language Selection to Local Storage

austinmatherne-wk:
#### Reason for change
User language selection for the application interface is lost on reload or when navigating between viewers.

#### Description of change
* Store user application language selection in local storage.
* During load, check if the application language was previously set in local storage.
* Explicit query param language still takes precedence.

#### Steps to Test
* Change application language.
* Refresh.
* Confirm application language persists.

**review**:
@Arelle/arelle
@paulwarren-wk


derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #822, opened 2025-02-12
Fall back to RTD user guide for Help > User Guide link

aaroncameron-wk:
#### Description of change
When the user guide feature is not enabled, fall back to the URL for the lastest ixbrl-viewer RTD documentation.

Add User Guide link to the header as well.

Note: I considered linking to the documentation version that matched the version of the viewer being used, but felt that wouldn't be best due to the user guide being in an early state right now.

#### Steps to Test
CI

image

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/util.js
+++ b/iXBRLViewerPlugin/viewer/src/js/util.js
@@ -32,6 +32,7 @@ export const STORAGE_APP_LANGUAGE = "ixbrl-viewer-app-language";
 export const STORAGE_THEME = "ixbrl-viewer-theme";
 export const STORAGE_HIGHLIGHT_FACTS = "ixbrl-viewer-highlight-all-facts";
 export const STORAGE_HOME_LINK_QUERY = "ixbrl-viewer-home-link-query";
+export const USER_GUIDE_URL = "https://arelle-ixbrl-viewer.readthedocs.io/";
 
 /* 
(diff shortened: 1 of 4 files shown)


austinmatherne-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:
FRC +1

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -32,6 +32,7 @@ export const STORAGE_APP_LANGUAGE = "ixbrl-viewer-app-language";
 export const STORAGE_THEME = "ixbrl-viewer-theme";
 export const STORAGE_HIGHLIGHT_FACTS = "ixbrl-viewer-highlight-all-facts";
 export const STORAGE_HOME_LINK_QUERY = "ixbrl-viewer-home-link-query";
+export const USER_GUIDE_URL = "https://arelle-ixbrl-viewer.readthedocs.io/";
```
brettkail-wk, line 35: Should this be https://arelle-ixbrl-viewer.readthedocs.io/en/latest/user_guides/user_guide.html (or even `__VERSION__`)?  It seems unfortunate to link to the root, which has all the building/installing/etc. information, which doesn't seem useful for most users.

PR #829, opened 2025-02-20
Enhancement: Use symbols in UTR for all units that have them defined (not just currencies)

strr:
#### Reason for change
I noticed that the measure labels (units) displayed in the viewer used the symbols from the UTR for currencies (so £ is displayed instead of GBP) but the symbols for other measures were not.

If you want to report an area using square meters, you need to use a measure of `utr:sqm` in your unit, which ixbrl-viewer will display as "sqm" with a tool tip of "Square metre". ixbrl-viewer should be displaying "m²" with a tool tip of "Square metre"

#### Description of change
Previously, currencies would use their symbol for their measure label but not other units. Shuffle the code around so that now all units will use their symbol if they have one defined and, if not, fall back to using their unit id as before.

In `utr.js`, which defines objects `UTR` and `UTREntry`, if a symbol was undefined for a given `UTREntry`, the name was used as the symbol. This didn't feel right, so I've moved the behaviour of "fallback on the name" to the only caller in `unit.js`: `measureLabel()` now handles all decisions about what and when to fallback to during the hunt for a suitable label for a measure.

Added a few tests to make sure this behaviour works and there are no regressions on the existing behaviour.
#### Steps to Test
Create an XBRL report that uses units from the UTR that have symbols that are different to their unitId (e.g. Fahrenheit, sqkm) and see that previously Arelle displayed the unitId and now it displays the symbol. The tool tip (name) is unchanged.

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/utr.js
+++ b/iXBRLViewerPlugin/viewer/src/js/utr.js
@@ -20,7 +20,7 @@ class UTR {
         if (u === undefined) {
             return undefined;
         }
-        return new UTREntry(u.s ?? u.n, u.n)
+        return new UTREntry(u.s ?? undefined, u.n)
     }
 }
 
--- a/iXBRLViewerPlugin/viewer/src/js/utr.js
+++ b/iXBRLViewerPlugin/viewer/src/js/utr.js
@@ -20,7 +20,7 @@ class UTR {
         if (u === undefined) {
             return undefined;
         }
-        return new UTREntry(u.s ?? undefined, u.n)
+        return new UTREntry(u.s, u.n)
     }
 }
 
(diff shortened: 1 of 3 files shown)


iXBRLViewerPlugin/viewer/src/js/utr.js
```diff
@@ -20,7 +20,7 @@ class UTR {
         if (u === undefined) {
             return undefined;
         }
-        return new UTREntry(u.s ?? u.n, u.n)
+        return new UTREntry(u.s ?? undefined, u.n)
```
brettkail-wk, line 23: Can the `?? undefined` be removed?

strr: Yes!

brettkail-wk (approved):

austinmatherne-wk (approved):

PR #955, opened 2025-11-12
Exclude non-essential files from Python and npm builds

austinmatherne-wk:
#### Reason for change
Similar to Arelle/Arelle/pull/1851 we're including files which aren't required in our Python and npm package builds. This is far less critical for ixbrl-viewer because 1. we don't release nearly as often and 2. the packages aren't as large. So while we're less likely to hit project size limits anytime soon, we still shouldn't bloat our artifacts unnecessarily.

#### Description of change
* Trim Python package down to only relevant files.
* Trim npm package down to only relevant files.
* Update Windows and macOS runner images to latest.
* Remove an old dependabot ignore setting. A newer version of dependency is already used.
* Update workflow to use latest LTS node version by default.

#### Steps to Test
- CI
- [x] Test [updated Python package](https://github.com/Arelle/ixbrl-viewer/actions/runs/19313573053)
- [x] Test [updated npm package](https://github.com/Arelle/ixbrl-viewer/actions/runs/19313576554)

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,10 +8,6 @@ updates:
       github-action-dependencies:
         patterns:
           - "*"
-    ignore:
-      - dependency-name: "softprops/action-gh-release"
-        versions:
-          - "2.2.0"
   - package-ecosystem: npm
     directory: "/"
     schedule:


derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1121, opened 2026-08-05
Anchor the fact tags to the concept title

austinmatherne-wk:
#### Reason for change

The target document and hidden fact tags were positioned absolutely against the inspector body.  The UI redesign left that rule alone while moving the markup: the body now opens with the previous/next nav, so the tags landed on top of it, covering the fact count and the Next button and reading as a chip floating over nothing.

#### Description of change
| master | pr |
| -- | -- |
| tags-before | tags-after |

#### Steps to Test

* [x] confirm fact tags render correctly

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/html/inspector.html
+++ b/iXBRLViewerPlugin/viewer/src/html/inspector.html
@@ -71,16 +71,18 @@ See COPYRIGHT.md for copyright information
                     </div>
                 </div>
               </nav>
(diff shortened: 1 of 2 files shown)


derekgengenbacher-wf (approved):

brettkail-wk (approved):

derekgengenbacher-wf:

PR #1136, opened 2026-08-06
Generate demo viewer on PRs for quick QA

aaroncameron-wk:
#### Reason for change
Make QA of frontend changes quicker by generating a demo viewer on each PR push.

#### Description of change
New demo workflow generates smoke test viewer.

#### Steps to Test
CI

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/.github/workflows/demo.yml
+++ b/.github/workflows/demo.yml
@@ -1,0 +1,48 @@
+name: Generate Demo Viewer
+
+on:
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  tests:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v7.0.1
+      - uses: actions/setup-node@v7.0.0
+        with:
+          cache: 'npm'
+          check-latest: true
+          node-version: '26'
+      - name: Install Python 3
+        uses: actions/setup-python@v7.0.0
+        with:
+          cache: 'pip'
+          cache-dependency-path: '**/pyproject.toml'
+          check-latest: true
+          python-version: '3.14'
+      - name: Install Dependencies
+        run: npm ci
+      - name: Install Arelle
+        run: pip install .[arelle]
+      - name: Build viewer js
+        env:
+          GIT_TAG: '0.0.0'
+        run: make prod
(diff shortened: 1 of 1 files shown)


.github/workflows/demo.yml
```diff
@@ -0,0 +1,42 @@
...
+
+permissions: {}
+
+jobs:
+  tests:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v7.0.1
+        with:
+          fetch-depth: 0 # npm run prod requires tags, sadly this is the only way to get them.
```
austinmatherne-wk: I think you can avoid this [by setting](https://github.com/Arelle/ixbrl-viewer/blob/50ed2d5d1a3fc07ebb9c062209aa02b23d6daff4/iXBRLViewerPlugin/viewer/version.js#L6) the `GIT_TAG` env var to a dummy value. Maybe worth it for this?

.github/workflows/demo.yml
```diff
@@ -0,0 +1,48 @@
...
+        env:
+          GIT_TAG: '0.0.0'
+        run: make prod
+      - name: Generate test viewers
+        shell: bash
+        run: |
+          mkdir demo
+          arelleCmdLine \
+            --plugins "ixbrl-viewer" \
+            -f "tests/puppeteer/test_filings/filing_documents_smoke_test.zip" \
```
austinmatherne-wk, line 42: This is already generated in the puppeteer workflow? Could that workflow be modified to produce artifacts that can be downloaded/demoed (probably need to drop `--viewer-no-copy-script` from the puppeteer script) instead of generating a viewer for the same report twice on each PR?

aaroncameron-wk: That was my initial approach, but I preferred not to update the puppeteer workflow to always publish artifacts or mess with the matrix properties to get only a single execution to upload an artifact. I also feel there's potential that we produce multiple or different viewers for demoing.

austinmatherne-wk: That's reasonable. I'm currently working on local changes that will 1. require more test docs and 2. replace puppeteer with playwright. I'll revisit this then

austinmatherne-wk (approved):

PR #1161, opened 2026-09-08
Wrap empty inline XBRL elements

### LINKED ISSUE #1015: [REQUEST] highlighting for facts with xsi:nil="true"

As more and more European filings seem to use `xsi:nil="true"` (probably accoring to the [ESEF reporting manual](https://www.esma.europa.eu/sites/default/files/library/esma32-60-254_esef_reporting_manual.pdf), Guidance 2.2.5), the iXBRL viewer should probably show these facts a bit more prominent.

E.g. if you look at this [filing](https://filings.xbrl.org/52990006IVXY7GCSSR39/2025-12-31/ESEF/FR/0/sartoriusstedimbio-2025-12-31-1-fr/reports/ixbrlviewer.html#f-fact-354) (fact 354) the NIL tagged value is almost invisible. To me it seems the viewer does not highlight the `td` (like for other facts) but creates an additional `<span>`, which spans "nothing" and due to that is not clickable nor highlighted.

Thanks and best regards,
Sacha

austinmatherne-wk:
#### Reason for change

Nil iXBRL fact values cannot currently be highlighted or selected in the DOM.

| master | branch |
| -- | -- |
| master  |  branch |

#### Description of change

Wrap empty elements, using the table cell when appropriate. Add unit tests for empty and nested elements.

#### Steps to Test

* CI

**review**:
@Arelle/arelle
@paulwarren-wk

### DIFF
--- a/iXBRLViewerPlugin/viewer/src/js/viewer.js
+++ b/iXBRLViewerPlugin/viewer/src/js/viewer.js
@@ -128,7 +128,9 @@ export class Viewer {
     // children, use those nodes as the wrappers.
     //
     // Otherwise, insert a wrapper node around the element.  If the node or any
-    // descendent has display: block, a div is used, otherwise a span.  
+    // descendent has display: block, a div is used, otherwise a span.  This
+    // includes elements with no content at all (e.g. nil facts), which would
+    // otherwise have no wrapper and so could not be highlighted or located.
     //
     // We want to avoid adding wrapper nodes around inline-block children, as
     // wrapping in block or inline-block can interfere with layout (e.g. some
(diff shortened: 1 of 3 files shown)


brettkail-wk (dismissed):

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -283,7 +288,7 @@ export class Viewer {
         const tableNode = domNode.closest("td,th");
         let nodes;
         const innerText = $(domNode).text();
-        if (tableNode !== null && getComputedStyle(tableNode).display === 'table-cell' && innerText.length > 0) {
+        if (tableNode !== null && getComputedStyle(tableNode).display === 'table-cell') {
```
brettkail-wk, line 294: I'm not entirely sure why commit 34abe4578fb6bfe8757b641d6163c7bcbaf81d0f added that `innerText.length > 0`, but the code appears to work correctly for empty string.

paulwarren-wk: It's not clear to me either. I suspect that it was to avoid calling `.indexOf("")` but behavior of this is well-defined, and the code does what we want in this case.

brettkail-wk (approved):

aaroncameron-wk (approved):

PR #1168, opened 2026-09-15
Batch style reads when creating fact wrappers

austinmatherne-wk:
#### Reason for change

Interleaving style reads with class changes caused repeated style recalculation. On [an Aviva report](https://ukixbrlviewer.org.uk/view/1a273a81-51cc-4dae-b30d-6e12b56cfd97/ixbrlviewer.html), batching the reads before writes reduced initial load from about 22 to 2 seconds. Other reports with heavy use of absolutely positioned content improved by a more modest, but still significant 1.5 to 2.5 seconds.

#### Description of change

Collect absolutely positioned descendants before applying wrapper classes. Preserve wrapper order and add unit and Puppeteer regression coverage.

#### Steps to Test

**review**:
@Arelle/arelle
@paulwarren-wk


derekgengenbacher-wf (dismissed):

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -305,19 +305,26 @@ export class Viewer {
         if (nodes === undefined) {
             nodes = this._wrapNode(domNode);
         }
-        const allNodes = [];
+        const subNodeLists = [];
```
brettkail-wk, line 309: Is it worth a small comment explaining why this is done in two passes so that someone doesn't think it should be cleaned up later?

brettkail-wk (dismissed):

aaroncameron-wk (approved):

brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1176, opened 2026-09-17
Speed up untagged number walk

austinmatherne-wk:
#### Reason for change

In review mode the viewer wraps every untagged number and date in a highlight span before the document becomes interactive to the user. On large filings this walk took several seconds. Most of that time went into rebuilding text nodes that had no match at all, and into hiding and showing the report body around the walk. The `postProcess` pass also interleaved layout reads and class writes, forcing a layout flush per element.

#### Description of change

The walk now uses plain DOM calls, iterates the child list in place, and leaves text nodes without a match untouched. The hide and show around the walk is removed, since the walk no longer forces layout. The `postProcess` pass collects zero height elements before adding the no highlight class so the reads no longer flush layout one at a time.

Benchmarking against large filings (Principal Funds, Pacific Life, and Blackstone Private Credit) in review mode, time to first user interaction was reduced by 8, 5, and 2 seconds. The wrapped markup is unchanged.

#### Steps to Test

1. Confirm load time is reduced in review mode.
2. Toggle the untagged numbers and untagged dates highlights and confirm the same numbers and dates are highlighted as on.

**review**:
@Arelle/arelle
@paulwarren-wk


iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -168,9 +168,10 @@ export class Viewer {
     }
 
 
-    _wrapUntaggedNumbers(n, docIndex, ignoreFullMatch) {
+    _wrapUntaggedNumbers(n) {
         const viewer = this;
         const ixHiddenStyleRE = /(?:^|\s|;)-(?:sec|esef)-ix-hidden:\s*([^\s;]+)/;
+        const ignoreFullMatch = localName(n.get(0).nodeName.toUpperCase()) === 'NONNUMERIC';
```
brettkail-wk: > The ignoreFullMatch flag was only ever set when recursing into a
> nonNumeric element, so it is a property of n rather than something
> every caller needs to thread through.

Doesn't this change semantics for `<ix:nonNumeric><span>...</span></ix:nonNumeric>`?  I'm not sure the current semantics make sense for `<ix:nonNumeric><span>...</span><span>...</span></ix:nonNumeric>` either, but that seems like a separate issue.

iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -169,13 +169,14 @@ export class Viewer {
...
 
     _wrapUntaggedNumbers(n) {
-        const viewer = this;
         const ixHiddenStyleRE = /(?:^|\s|;)-(?:sec|esef)-ix-hidden:\s*([^\s;]+)/;
         const ignoreFullMatch = localName(n.nodeName.toUpperCase()) === 'NONNUMERIC';
 
-        $(n).contents().each(function () {
-            if (this.nodeType === Node.ELEMENT_NODE) {
-                const name = localName(this.nodeName.toUpperCase());
+        // Capture the next sibling first because wrapping replaces text nodes.
```
brettkail-wk: This comment might be clearer on the next line?

iXBRLViewerPlugin/viewer/src/js/viewer.js
austinmatherne-wk: > Doesn't this change semantics for \<ix:nonNumeric>\<span>...\</span></ix:nonNumeric>?

On master the flag is passed fresh as `name === 'NONNUMERIC'` on every recursive call, so it only ever described the immediate parent, which is what the new line computes from `n`. The nested span case wraps on both master and this branch, and I agree that is worth a separate issue. I'd prefer to handle that in another PR and not mix refactor/perf optimizations with behavioral changes.

brettkail-wk: Ah, I see, thanks.

brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1177, opened 2026-09-17
Treat a nonNumeric with a single value as tagged

austinmatherne-wk:
#### Reason for change

A `nonNumeric` whose value sits inside formatting markup, such as `<ix:nonNumeric><span>2020</span></ix:nonNumeric>`, is highlighted as untagged in review mode, as raised in https://github.com/Arelle/ixbrl-viewer/pull/1176#discussion_r4036951006.

#### Description of change

The walk now checks the whole text content of each format free `nonNumeric` once with an anchored matcher and skips it as tagged when the trimmed text is a single number or date. Nested `nonNumeric`s are checked on their own.

Performance was checked against #1176 to verify no regressions.

#### Steps to Test

* CI

**review**:
@Arelle/arelle
@paulwarren-wk


iXBRLViewerPlugin/viewer/src/js/viewer.js
```diff
@@ -180,36 +179,31 @@ export class Viewer {
...
         const input = node.nodeValue;
         const doc = node.ownerDocument;
         let output = null;
         let pos = 0;
         numberMatchSearch(input, (m, do_not_want, is_date) => {
-            // A match covering the whole of a nonNumeric's text content
-            // is considered tagged.
-            if (do_not_want ||
-                    (ignoreFullMatch && m.index === 0 && m.index + m[0].length === input.length && input === node.parentNode.textContent)) {
+            if (do_not_want) {
```
brettkail-wk: This is fine, but at this point, can `numberMatchSearch` just not invoke the callback for `do_not_want` (and remove that param from the callback so that's all hidden in that module)?

brettkail-wk (dismissed):

brettkail-wk (approved):

derekgengenbacher-wf (approved):

derekgengenbacher-wf:

PR #1178, opened 2026-09-18
Use MessageChannel for generator tasks

austinmatherne-wk:
#### Reason for change

Browsers clamp deeply nested timers to a [4 ms delay](https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#nested_timeouts). This costs roughly 150-300 ms on standard SEC filings and 3-7 seconds on larger Principal, Pacific, and Blackstone reports.

> As specified in the [HTML standard](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers), browsers will enforce a minimum timeout of 4 milliseconds once a nested call to setTimeout has been scheduled 5 times.

#### Description of change

Use `MessageChannel` for work that runs after the report becomes visible (building the search index and checking which elements can be highlighted).

#### Steps to Test

* CI

**review**:
@Arelle/arelle
@paulwarren-wk


derekgengenbacher-wf (dismissed):

iXBRLViewerPlugin/viewer/src/js/util.js
```diff
@@ -192,17 +192,22 @@ export function setDefault(obj, key, def) {
 }
 
 export function runGenerator(generator, onDone) {
-    function resume() {
+    // MessageChannel avoids the 4 ms minimum delay imposed on nested timers.
```
brettkail-wk: This is completely fine.  This comment makes sense in the context of the PR, but it seems to make less sense in isolation.
> Use MessageChannel rather than setTimeout to implement yielding since browsers impose a 4 ms minimum delay on nested timers.

Or a function-level comment might help:
> Run the generator to completion, yielding to the event loop after each iteration.

brettkail-wk (dismissed):

iXBRLViewerPlugin/viewer/src/js/util.js
austinmatherne-wk: Updated

derekgengenbacher-wf (approved):

brettkail-wk (approved):

derekgengenbacher-wf:

